PR #38, opened 2022-04-04
Add prod config

estellecomment:
Not necessarily perfect, it's based on the preprod config, which we haven't cleaned up and understood fully yet.


celiacheff (approved):

estellecomment:
Based on @celiacheff 's suggestion, I also removed registration which doesn't work.
And added threads ! 🧶

PR #39, opened 2022-04-08
Fix node version to ~~v16.13.2~~ 14

estellecomment:
Fixing the node version, so that we standardise the environment in all our deployments.
Using v16.13.2 (latest v16) because with v17 I get `ERR_OSSL_EVP_UNSUPPORTED` error : 
```
node:internal/crypto/hash:67
this[kHandle] = new _Hash(algorithm, xofLen);
Error: error:0308010C:digital envelope routines::unsupported
```

---
This PR currently has no changelog labels, so will not be included in changelogs.

Add one of: `T-Deprecation`, `T-Enhancement`, `T-Defect`, `T-Task` to indicate what type of change this is plus `X-Breaking-Change` if it's a breaking change.


celiacheff (approved):
I had the same problem too, but didn't check what are Element's plans on this topic.

estellecomment:
Hmm good question. 
`node-version` file says 14, and a coupe other mentions as well. 
Asking element why they are at 14, so that we can make an informed decision.

odelcroi:
Docker image uses node-14
`FROM node:14-buster as builder`

but recommendation from readme says "Ensure you have the latest LTS version of Node.js installed."

I guess we can cope with `"node": ">=14.19"`

estellecomment:
Scalingo build now says :
```
-----> Installing binaries
       engines.node (package.json):  >=14.19 <15
       engines.npm (package.json):   unspecified (use default)
       engines.yarn (package.json):  unspecified (use default)
       
       Resolving node version >=14.19 <15...
       Downloading and installing node 14.19.3...
       Using default npm version: 6.14.17
       Resolving yarn version 1.22.x...
       Downloading and installing yarn (1.22.19)
       Installed yarn 1.22.19
```
Looks good !

PR #59, opened 2022-04-25
[refactor] Get room federation options from config

estellecomment:
Hard-coding which instances has specific behavior for room federation is a problem : you cannot test outside of the hard-coded prod instances.
Make room federation options be fetchable from config, with prod or preprod servers.

---
This PR currently has no changelog labels, so will not be included in changelogs.

Add one of: `T-Deprecation`, `T-Enhancement`, `T-Defect`, `T-Task` to indicate what type of change this is plus `X-Breaking-Change` if it's a breaking change.


estellecomment:
closing, this is really old

PR #62, opened 2022-06-13
Rebase/element web 1.10.14

estellecomment:
Catch up with element-web 1.10.14
Note that this is actually a git merge, not a git rebase. I'm using "rebase" because it's the word the mobile devs use.

Important things in this rebase : 
 - reskindex has been removed. This removes build steps, the build becomes simpler. It also removes the need for sdk.getComponent in our customisations, we can now use normal imports.
  - replaceableComponent has been removed, we will now use customisations instead (we had already started)


estellecomment:
Checks are broken, filed issue : https://github.com/tchapgouv/tchap-web-v4/issues/65
Not going to fix now.

jdauphant:
We didn't have auto-deploy on Tchap Web V4 ?

estellecomment:
@jdauphant we have review apps, it was deleted because the PR was too old, I just restarted it

celiacheff (approved):
Ok pour moi, bonne maj, faudra probablement désactiver les workflow GH qu'on utilise pas.

PR #69, opened 2022-06-27
Fix space-panel color to be different from the button color, to be able to see the space buttons

estellecomment:
Before : can't see the buttons to add a space or view a sub space : 
Screen Shot 2022-06-27 at 15 50 29

After : see the buttons !
Screen Shot 2022-06-27 at 15 50 20

### DIFF
--- a/config.dev.json
+++ b/config.dev.json
@@ -76,7 +76,7 @@
                     "private-color": "#eb5757",
                     "external-color": "#f07a12",
                     "forum-color": "#27ae60",
-                    "sidebar-color": "#27303a",
+                    "sidebar-color": "#c0d8ef",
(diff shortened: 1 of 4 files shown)


celiacheff (approved):

celiacheff:
Do we also have a dark theme ?

estellecomment:
> Do we also have a dark theme ?

@celiacheff no we don't. A good idea for later :)

jdauphant:
Maintening 4 files looks complicated :thinking: 
Can we simplify ? (maybe we can merge json)

estellecomment:
Yes, I agree it's not great. We need to work on it with infra

PR #131, opened 2022-08-12
Add MCS API

weeman1337:
---
This change is marked as an *internal change* (Task), so will not be included in the changelog.


src/content-scanner/ContentScanner.ts
```diff
@@ -0,0 +1,140 @@
+/*
```
estellecomment, line 1: not sure what the legal situation is, should we be having copyrighted New Vector code in tchap code ?

src/content-scanner/readScannerUrlConfig.ts
```diff
@@ -0,0 +1,34 @@
...
+import SdkConfig from "matrix-react-sdk/src/SdkConfig";
+
+import { ContentScannerConfig } from "./ContentScannerConfig";
+
+/**
+ * Reads "content_scanner.url" from the config.
+ * Logs a warning if it is not present.
+ */
+export const readScannerUrlConfig = (): string => {
+    const contentScannerUrl = (SdkConfig.get() as unknown as ContentScannerConfig)?.content_scanner?.url;
```
jdauphant: On Tchap the content scanner is the base as the server. We have 17 differents homeservers different

src/content-scanner/readScannerUrlConfig.ts
```diff
@@ -0,0 +1,34 @@
+/*
```
jdauphant: This file start by lower case but other file start by upper case

src/content-scanner/readScannerUrlConfig.ts
jdauphant: The implementation should use the current homeserver of the user as content scanner endpoint and not an configurated value

src/content-scanner/ContentScanner.ts
```diff
@@ -0,0 +1,140 @@
...
+
+/**
+ * Content scanner implementation that interacts with a Matrix Content Scanner.
+ * @see https://github.com/matrix-org/matrix-content-scanner
+ */
+export class ContentScanner implements ContentScannerInterface {
+    private static internalInstance: ContentScanner;
+
+    private mcsKey: PkEncryption = new global.Olm.PkEncryption();
+    private keySet = false;
```
estellecomment: "keySet" sounds like a set of keys. "hasKey" would be clearer to me.

src/content-scanner/ContentScanner.ts
giomfo: @estellecomment we are using this header in all our contributions in Tchap mobile fork. It should be the same here. Let's talk in DM if need

src/content-scanner/readScannerUrlConfig.ts
weeman1337: Yes. This file only exports a function with that name `readScannerUrlConfig` while the other files export components or classes like for instance `ContentScanner`.

This file is going to be removed anyway.

src/content-scanner/readScannerUrlConfig.ts
weeman1337: Going to use the matrix server URL here.

src/content-scanner/ContentScanner.ts
weeman1337: Done

weeman1337:
Closed in favour of #132

Issue #130, opened 2022-07-26
Improve Theme customization


jdauphant:
#### What would you like to do?
Having a way to customize CSS

#### Why would you like to do it?
To improve the design.

#### How would you like to achieve it?
I want to override https://github.com/matrix-org/matrix-react-sdk/blob/develop/res/themes/dark/css/dark.pcss
https://github.com/matrix-org/matrix-react-sdk/blob/develop/res/themes/light/css/light.pcs

And define our own css variables and fonts

jdauphant:
@estellecomment @odelcroi cette PR rend autonome @areox-net pour modifier le théme est améliorer l'expérience.
On pourra en discuter pour que vous faites un avis technique sur le projet

odelcroi:
Definitely interesting, let's investigate it after the beta

odelcroi:
linked to #186, @jdauphant have you something in mind?

odelcroi:
relates to #273

estellecomment:
We now have tchap pcss files which we can use to define our themes. I think this fixes this issue. (reopen if not)

jdauphant:
@estellecomment can we link the PR to this ticket ?

jdauphant:
Found it !
Closed by #277

PR #132, opened 2022-08-17
Add MCS

weeman1337:
---
This change is marked as an *internal change* (Task), so will not be included in the changelog.


src/components/views/messages/OriginalImageBody.tsx
```diff
@@ -0,0 +1,612 @@
...
+import { MatrixClientPeg } from 'matrix-react-sdk/src/MatrixClientPeg';
+import { presentableTextForFile } from 'matrix-react-sdk/src/utils/FileUtils';
+import { ImageSize, suggestedSize as suggestedImageSize } from 'matrix-react-sdk/src/settings/enums/ImageSize';
+import Spinner from 'matrix-react-sdk/src/components/views/elements/Spinner';
+import warningIcon from "matrix-react-sdk/res/img/warning.svg";
+
+import MFileBody from "./OriginalFileBody";
+
+/**
+ * Original MImageBody from react-sdk.
```
estellecomment: can you specify the version of react-sdk, for tracking purposes ?
Currently tchap-web-v4 uses react-sdk 3.46.0, which has a slightly different version of this file.

jdauphant (commented):
Solution to not having to copy manually the original files :
- Find a way to reference original files from the react sdk
- Create symlinks to original files
- Having a pre-compile script that auto-copy the original files before compilation

And if can't : we can add a CI check that Original is always the same.

jdauphant:
> Solution to not having to copy manually the original files :
> 
> * Find a way to reference original files from the react sdk
> * Create symlinks to original files
> * Having a pre-compile script that auto-copy the original files before compilation
> 
> And if can't : we can add a CI check that Original is always the same.

Another solution have been suggested by the beta.gouv.fr communauty : https://github.com/ds300/patch-package

src/customisations/components/views/messages/ContentScanningAudioBody.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+    public render() {
+        if (this.state.isScanning || !this.state.isSafe) {
+            return <ContentScanningFileBody {...this.props} />;
+        }
+
+        return <div>
+            <OriginalAudioBody {...this.props} />
+            <ContentScanningStatus status="done" />
+        </div>;
```
estellecomment, line 73: This displays "Scan done" even when there is no scan (the NoopContentScanner has been used). That's a bit misleading to me, the user could get a false sense of security.

estellecomment (commented):
When there is infected content, we should help the user understand what they can or can't do. They might get scared that the content will infect their machine for example. I think we need a bit more help text. For example in a dialog that opens when you try to download, and also when you hover on the red bug icon.
Has this feature been user-tested ?
All that can be done in a separate PR though, it's an enhancement, this version does the job.

src/i18n/strings/en_EN.json
```diff
@@ -15,6 +15,43 @@
...
+    "Error": "Error",
+    "Error decrypting attachment": "Error decrypting attachment",
+    "Decrypt %(text)s": "Decrypt %(text)s",
+    "Invalid file%(extra)s": "Invalid file%(extra)s",
+    "Error decrypting image": "Error decrypting image",
+    "Show image": "Show image",
+    "Error decrypting video": "Error decrypting video",
+    "Error processing voice message": "Error processing voice message",
+    "Scanning": "Scanning",
+    "Scan done": "Scan done",
```
estellecomment: The text we have on the android client is "Trusted", let's reuse the same.
We also need a french translation, "Vérifié"

estellecomment: We avoid editing this file, and instead we add tchap-specific translations to src/i18n/strings/tchap_translations.json
It's easier to track that way.

src/components/views/messages/OriginalImageBody.tsx
weeman1337: @estellecomment can you check whether https://github.com/tchapgouv/tchap-web-v4/pull/132/commits/1042876677a946e2d0424d96a21de4d46a346a1b works for you?

This is a quite creative solution based on your comments about using `customisations.json` to point to `node_modules` and Juliens hint to `patch-package`.

package.json
```diff
@@ -25,9 +25,6 @@
     "package.json",
     "contribute.json"
   ],
-  "engines": {
```
weeman1337: Commited by mistake. Will revert this.

src/i18n/strings/en_EN.json
weeman1337: It is now „Trusted“/„Vérifié“ . Moved the translations to `tchap_translations.json`.

src/customisations/components/views/messages/ContentScanningAudioBody.tsx
weeman1337: Dropped the `NoopContentScanner`.

src/content-scanner/ContentScanner.ts
```diff
@@ -88,29 +87,44 @@ export class ContentScanner {
...
-                        .then(r => r.json());
-                    this.mcsKey.set_recipient_key(k["public_key"]);
-                    this.hasKey = true;
+                    try {
+                        const k = await fetch(this.scannerUrl + "/_matrix/media_proxy/unstable/public_key")
+                            .then(r => r.json());
+                        this.mcsKey.set_recipient_key(k["public_key"]);
+                        this.hasKey = true;
+                    } catch (err) {
+                        reject(err);
```
estellecomment: does catch + reject actually do anything ? Wouldn't throwing in a promise cause the promise to be rejected anyway ?

src/customisations/components/views/messages/ContentScanningDownloadActionButton.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+        let tooltip = _t("Scanning");
+
+        if (!this.state.isScanning) {
+            if (this.state.isSafe) {
+                tooltip = spinner ? _t("Decrypting") : _t("Download");
+            } else if (this.state.hasError) {
+                tooltip = _t("Scan unavailable");
+            } else if (!this.state.isSafe) {
+                tooltip = _t("Infected content");
```
estellecomment: this is "Content Blocked" now

src/content-scanner/ContentScanner.ts
weeman1337: Throwing in a Promise directly yes. But here the errors are raised in other async functions. That would lead to an uncaught in promise error.

weeman1337: Doing everything in an `async` function works: https://github.com/tchapgouv/tchap-web-v4/pull/132/commits/b5902acfe2bcd30b8c715acb9dd0bef5e7cc12f3

src/content-scanner/ContentScanner.ts
```diff
@@ -86,52 +86,38 @@ export class ContentScanner {
...
+        const prom = this.doScan(mxc, file);
+        this.pendingScans.set(mxc, prom);
+        return prom;
+    }
+
+    private async doScan(mxc: string, file?: IEncryptedFile): Promise<boolean> {
+        let response: Response;
+
+        if (file) {
+            if (!this.hasKey) {
```
estellecomment, line 104: the key fetching could be factored out into a named function

src/content-scanner/ContentScanner.ts
```diff
@@ -86,52 +86,38 @@ export class ContentScanner {
...
-            } else {
-                const url = this.scannerUrl + `/_matrix/media_proxy/unstable/scan/${mxc.substring('mxc://'.length)}`;
-                try {
-                    response = await fetch(url);
-                } catch (err) {
-                    reject(err);
-                    return;
-                }
+        const prom = this.doScan(mxc, file);
+        this.pendingScans.set(mxc, prom);
```
estellecomment: "pendingScans" is a bit inaccurate : it also contains finished scans. (I just realized that finished scans are not removed, so this doesn't do what I thought!)

Maybe "startedScans" or "alreadyScanned" or "cachedScans"?

src/customisations/components/views/messages/ContentScanningDownloadActionButton.tsx
weeman1337: Done.

src/content-scanner/ContentScanner.ts
weeman1337: Renamed to `cachedScans`

src/content-scanner/ContentScanner.ts
weeman1337: Done ✓

src/content-scanner/ContentScanner.ts
```diff
@@ -123,6 +123,7 @@ export class ContentScanner {
     public static get instance(): ContentScanner {
         if (!ContentScanner.internalInstance) {
             ContentScanner.internalInstance = new ContentScanner(MatrixClientPeg.get().getHomeserverUrl());
+            ContentScanner.internalInstance = new ContentScanner("http://localhost:9000");
```
estellecomment: accidental commit ?

src/content-scanner/ContentScanner.ts
```diff
@@ -120,6 +117,13 @@ export class ContentScanner {
         return responseJson.clean;
     }
 
+    private async fetchKey() {
```
estellecomment, line 126: can you also use this function in the `download` function ? (l. 66)

src/customisations/components/views/messages/ContentScanningDownloadActionButton.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+    public constructor(props: IProps) {
+        super(props);
+
+        this.state = {
+            hasError: false,
+            isSafe: false,
+            isScanning: false,
+            loading: false,
+            tooltip: _td("Downloading"),
```
estellecomment: not sure this is the right init value. "Download" would make more sense, or just leave `render` to figure it out ?

src/content-scanner/ContentScanner.ts
weeman1337: Yes, of course. This is sub-optimal. I have to change this every time I am developing locally. Maybe I re-add the URL option for the content scanner. This way I can override it locally. What do you think?

src/customisations/components/views/messages/ContentScanningDownloadActionButton.tsx
weeman1337: This state property can be removed anyway.

src/content-scanner/ContentScanner.ts
estellecomment: hmm yes that sucks.
Re-add the `content_scanner.url`, and if it's absent, default to `homeserverUrl` ? That way we have prod config by default, and no possibility of running without contentscanner, that sounds good.

weeman1337: Done

src/customisations/components/views/messages/ContentScanningDownloadActionButton.tsx
```diff
@@ -0,0 +1,161 @@
...
+import { MediaEventHelper } from "matrix-react-sdk/src/utils/MediaEventHelper";
+import { FileDownloader } from "matrix-react-sdk/src/utils/FileDownloader";
+import { _t } from "matrix-react-sdk/src/languageHandler";
+import Spinner from "matrix-react-sdk/src/components/views/elements/Spinner";
+import { RovingAccessibleTooltipButton } from "matrix-react-sdk/src/accessibility/RovingTabIndex";
+
+import { Media } from "../../../ContentScanningMedia";
+import { BlockedIcon } from "../../../../components/views/elements/BlockedIcon";
+
+enum DownloadState {
```
estellecomment, line 30: much cleaner, nice job !
The bug is gone, confirmed by testing by hand.

package.json
```diff
@@ -149,6 +151,8 @@
     "postcss-preset-env": "^6.7.0",
     "postcss-scss": "^2.1.1",
     "postcss-simple-vars": "^5.0.2",
+    "postcss-strip-inline-comments": "^0.1.5",
```
estellecomment: did we add this for a reason ?

res/css/views/elements/BlockedIcon.pcss
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+.mx_BlockedIcon {
+    background-color: #f07a12;
```
estellecomment, line 18: Would be nice to reuse colors, like --var(warning) or something.
Not a blocker, if it's not a 5 minute task then let's file an issue instead

estellecomment (approved):
AND THIS IS DONNNNNNNE
Congrats everyone, I would like to thank my parents without whom I would not have made it so far

(two small comments)

package.json
weeman1337: Leftover from merge… Removed

res/css/views/elements/BlockedIcon.pcss
weeman1337: Totally agree. Unfortunately, when not using a custom theme we don't have this variables. This is a > 5 minutes task.

jdauphant: WIth Angelo, we want to use a different way to custom the theme : https://github.com/tchapgouv/tchap-web-v4/issues/130 
It could be add to the issue when it will be priorized.

jdauphant:
Are we good to merge this PR ?

estellecomment:
Yes, good for me, @weeman1337 are you still working on some things ?

PR #161, opened 2022-09-08
111 - rm spaces option if spaces creation is disabled

### LINKED ISSUE #111: Investigate : Room options allows changing public/private status of room
edit : work to do

### What v2 options are
Rooms version : 1
for a private room : 
- Autoriser l’accès aux externes à ce salon  
- Activer l'accès au salon par lien 

for a private room with extern : 
- Activer l'accès au salon par lien 
- [activated] Autoriser l’accès aux externes à ce salon  (can not be changed)

for public rooms (forum : 
-  Retirer ce salon de l'annuaire des forums
- [activated] Activer l'accès au salon par lien (can not be changed)

### What should be done in v4 to stick to v2 behaviour
Rooms version : 1
for a private room remove the possibility to change : 
- [ ] Publier ce salon dans le répertoire de salons public (make no sense)
- [x] Accès> Membres de l’espace
- [x] Accès> Public

for a private room with externe remove the possibility to change  : 
- [ ] Publier ce salon dans le répertoire de salons public (make no sense)
- [x] Accès> Membres de l’espace
- [x] Accès> Public

for a public room remove the possibility to change : 
- [x] Chiffré (public room must not be not encrypted)
- [x] Accès> Membres de l’espace
- [x] Accès> Private
- [ ] Activer l’accès visiteur

### What should be added in v4 

for a private room
- [x] #169

----

Questions : 
- should be continue creating rooms with version 1 or upgrade to new room version? 

Initial ticket : 
In tartine, you can change a public room to private, and back again. 
In Tchap-style rooms, there are several settings that go together to define the type of the room. Not sure what happens if you change only public/private.

Capture d’écran 2022-07-15 à 14 32 48

odelcroi:
For all 3 types of room : 

Before : 

Capture d’écran 2022-09-08 à 11 22 38

After : 

Capture d’écran 2022-09-08 à 11 18 48

## Checklist

* [ ] Tests written for new code (and old code if feasible)
* [ ] Linter and other CI checks pass
* [ ] Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/vector-im/element-web/blob/develop/CONTRIBUTING.md))


src/components/views/settings/TchapJoinRuleSettings.tsx
```diff
@@ -0,0 +1,330 @@
...
+                if (!Array.isArray(restrictedAllowRoomIds)) return;
+                if (restrictedAllowRoomIds.length > 0) {
+                    onRestrictedRoomIdsChange(restrictedAllowRoomIds);
+                } else {
+                    onChange(JoinRule.Invite);
+                }
+            };
+
+            /**
+             * TCHAP : disable space-related options if create space feature is not enabled
```
estellecomment, line 162: Hmmm, we are losing future evolutions of element's JoinRuleSettings.tsx... 🤔 Maybe @jdauphant 's idea of using CSS to hide things would be easier to maintain... 
But this works and moves forward. We can try the CSS way in another PR.

The e2e test is the same in both cases (yay e2e test !)

src/components/views/settings/TchapJoinRuleSettings.tsx
```diff
@@ -0,0 +1,330 @@
...
+             */
+            if (ComponentVisibilityCustomisations.shouldShowComponent(UIComponent.CreateSpaces)) {
```
estellecomment, line 164: I wish we could contribute this to element-web, but it makes sense only for a project without spaces, which element doesn't allow. 
(in the element-web situation, if spaces creation is disabled, the user can still have spaces, and chose the visibility of the room in these spaces.)

src/components/views/settings/TchapJoinRuleSettings.tsx
estellecomment: Other idea : instead of customisations.json, use patch-package to add the `if` into element's JoinRuleSettings.tsx. 

It would not be too hard to maintain : at each upgrade of react-sdk, `yarn install` triggers a patch-package warning, like "the patch was made for vX.X.X and you are now at v.Y.Y.Y, no garantees". And if the change is too great, the patching breaks and yarn install fails.

And the e2e test is STILL the same !

odelcroi:
Duplicate of #166

PR #165, opened 2022-09-12
E2e tests without dockerized synapse

estellecomment:
Tchap doesn't have a dockerized synapse. So run e2e tests against the dev instance. 
This PR keeps the docker-synapse code around for later, but doesn't use it any more.
Also fixes language problems, and some other cleanup.

Still doesn't run in CI. Yet. :)


odelcroi (approved):

odelcroi:
looks good, havent tested it locally, will soon!

PR #168, opened 2022-09-14
E2e tests : add logout

estellecomment:
So that tests can clean up after themselves.


estellecomment:
Actually, not ready for review. 
I need to understand how chaining of commands work in cypress, because things are not being run when I thought.

estellecomment:
Cookies are deleted after each test by default by Cypress. This includes session cookies. 
https://docs.cypress.io/api/cypress-api/cookies

So logging out explicitly is not needed. 
(Conversely, the link above suggests ways to STAY logged in between tests to avoid spending too much time logging in at each test)

PR #172, opened 2022-09-14
Allow external users from private room

### LINKED ISSUE #169: Settings : Authorize access to external from a private room
In the Room Settings , Security and Privacy, we would like to add the feature to allow the access to external users from a private room for the administrator

Before: 
image

After :
image
- When clicking on the toggle : 
image

### Dev notes
- this feature is using the following synapse module : https://github.com/vector-im/synapse-room-access-rules
- code from matrix-react-js for this feature(when changing this particular setup) : https://github.com/tchapgouv/matrix-react-sdk/commit/1bdada20ef7fcd701c338984bda4d5aba20d76ea#diff-9f226359a8fc92d20a9572db5ad4ad640dbc337b2e96d25f163ae988ecf43b20
- code from matrix-react-js for this feature(when creating a room) : https://github.com/tchapgouv/matrix-react-sdk/commit/f22ec41d4982ba95bf4e1c711bf5268bb397b85d

### Task to perform to resolve this ticket : 
- [x] display the toggle switch 
- [x] perform api call to backend in order to allow the access to externals
- [x] add the following guards when allowing the access to externals : 
  - [x] role = administrator of the room
  - [x] room state = private
  - [x] accessRule = restricted
- [x] add a popup window to confirm if we want to change the access to externals
  - Title of the popup : Allow the externals to join this room
  - Description of the popup : This action is irreversible. Are you sure you want to allow the externals to join this room ?
- [ ] Add a description/information/label on the room `Room opened to external users` if the room is opened to external users
- [x] perform api call to backend when creating a private room opened to external users
- [ ] we cannot leave a room that is opened to external : is that normal?
- [x] add e2e tests

mcalinghee:
### Allow external users from private room
Screenshot 2022-09-27 at 16 29 04

### Create a private room opened to external users(bug fix)
This PR allow external users from a private room but also create a private room opened to external users(fix a small bug)

**Before:**
1. Create a room opened to externals 
Screenshot 2022-09-27 at 16 38 53
2. Settings in v2 : we expect that the room is opened to external => this is not the case here
Screenshot 2022-09-27 at 16 41 33
3. Settings of the room in v4 : we expect that the room is opened to external => this is not the case here
Screenshot 2022-09-27 at 16 44 58

**After:**
1. Create a room opened to externals 
image
2. Settings in v2 : we expect that the room is opened to external => OK
Screenshot 2022-09-27 at 16 36 47
3. Settings of the room in v4 : we expect that the room is opened to external => OK
image


src/components/views/settings/TchapJoinRuleSettings.tsx
```diff
@@ -105,10 +114,56 @@ const JoinRuleSettings = ({ room, promptUpgrade, aliasWarning, onError, beforeCh
...
+        //             }
+        //         },
+        //     });
+        // };
+        const onExternalAccessChange = async () => {
+            setContentAccess({ "rule": TchapRoomAccessRule.Unrestricted });
+        };
+
+        const openedToExternalUsers = accessRule.rule === TchapRoomAccessRule.Unrestricted;
+        const privateDescription = <div>
```
mcalinghee: rename to a more readable name : privateDescription

src/lib/createTchapRoom.ts
```diff
@@ -23,41 +18,61 @@ export default class TchapCreateRoom {
...
         createRoomOpts.creation_content = { 'm.federate': federate };
+        createRoomOpts.initial_state = createRoomOpts.initial_state || [];
 
         switch (tchapRoomType) {
             case TchapRoomType.Forum: {
             //"Forum" only for tchap members and not encrypted
-                createRoomOpts.accessRule = TchapRoomAccessRule.Restricted;
                 createRoomOpts.visibility = Visibility.Public;
                 createRoomOpts.preset = Preset.PublicChat;
+                createRoomOpts.initial_state.push({
```
mcalinghee, line 40: To set the access of a room, I have decided to use `createRoomOpts .initial_state` here instead of `createRoomOpts. accessRule` as we would need to :
- change the code in createRoom.ts and create tchapCreateRoom.ts (which should be ok)
- but we need also to change the code in `room-utils.ts` in particular :
```ts
return cy.createRoom(TchapCreateRoom.roomCreateOptions(roomName, TchapRoomType.External, false).createOpts);
```
This code is not taking into account the access rule settings and does not push the event to change the access rule therefore this part would need to change as well

src/components/views/settings/TchapJoinRuleSettings.tsx
```diff
@@ -63,12 +66,18 @@ const JoinRuleSettings = ({ room, promptUpgrade, aliasWarning, onError, beforeCh
...
     const restrictedAllowRoomIds = joinRule === JoinRule.Restricted
         ? content.allow?.filter(o => o.type === RestrictedAllowType.RoomMembership).map(o => o.room_id)
         : undefined;
 
+    const [contentAccessRule, setContentAccess] = useLocalEcho<IAccessRuleEventContent>(
+        () => room.currentState.getStateEvents(RoomAccessRulesEventId, "")?.getContent(),
+        content => cli.sendStateEvent(room.roomId, RoomAccessRulesEventId, content, ""),
+        onError,
+    );
+    const { rule: accessRule = undefined } = contentAccessRule || {};
```
estellecomment: pourquoi le default undefined, sachant qu'on a deja `|| {}` ?

src/components/views/settings/TchapJoinRuleSettings.tsx
```diff
@@ -105,10 +114,42 @@ const JoinRuleSettings = ({ room, promptUpgrade, aliasWarning, onError, beforeCh
     const definitions: IDefinition<JoinRule>[] = [];
 
     if (joinRule === JoinRule.Invite) {
```
estellecomment, line 117: Is this selecting private rooms only ? So far we were are copying android's implementation for determining what type of room it is : 
https://github.com/tchapgouv/tchap-android/blob/develop/vector/src/main/java/fr/gouv/tchap/core/utils/RoomUtils.kt#L31
(selection on isEncrypted and accessRules)

It would be nice to have a utils function (in TchapUtils for example) that checks what type of room it is.
v2 had a [isRoomForum function](https://github.com/tchapgouv/matrix-react-sdk/blob/26d2e9396b5e0a01cc8973b3aa7b09ce94f5a7da/src/Tchap.js#L194-L195) for example, we could do something similar

src/i18n/strings/en_EN.json
```diff
@@ -28,5 +28,8 @@
     "Decentralised, encrypted chat &amp; collaboration powered by $matrixLogo": "Decentralised, encrypted chat &amp; collaboration powered by $matrixLogo",
     "Sign In": "Sign In",
     "Create Account": "Create Account",
-    "Explore rooms": "Explore rooms"
+    "Explore rooms": "Explore rooms",
+    "Allow the externals to join this room": "Allow the externals to join this room",
```
estellecomment:
```suggestion
    "Allow external users to join this room": "Allow external users to join this room",
```

src/i18n/strings/en_EN.json
```diff
@@ -28,5 +28,8 @@
     "Decentralised, encrypted chat &amp; collaboration powered by $matrixLogo": "Decentralised, encrypted chat &amp; collaboration powered by $matrixLogo",
     "Sign In": "Sign In",
     "Create Account": "Create Account",
-    "Explore rooms": "Explore rooms"
+    "Explore rooms": "Explore rooms",
```
estellecomment: please add tchap-specific strings in tchap_translations.json, rather than the element files, so we can keep track of them.

src/@types/tchap.ts
```diff
@@ -9,3 +9,9 @@ export enum TchapRoomAccessRule {
   Unrestricted = "unrestricted", // accessible to externals
   Restricted = "restricted" // not accessible to externals
 }
+
+export interface IAccessRuleEventContent {
```
estellecomment: Can we use "tchap" in names with AccessRule ? It's a custom Tchap event.
(In this file, we have half of names with "tchap" and the other half without)

src/components/views/settings/TchapJoinRuleSettings.tsx
```diff
@@ -63,12 +66,18 @@ const JoinRuleSettings = ({ room, promptUpgrade, aliasWarning, onError, beforeCh
         content => cli.sendStateEvent(room.roomId, EventType.RoomJoinRules, content, ""),
         onError,
     );
-
     const { join_rule: joinRule = JoinRule.Invite } = content || {};
     const restrictedAllowRoomIds = joinRule === JoinRule.Restricted
         ? content.allow?.filter(o => o.type === RestrictedAllowType.RoomMembership).map(o => o.room_id)
         : undefined;
 
+    const [contentAccessRule, setContentAccess] = useLocalEcho<IAccessRuleEventContent>(
```
estellecomment: I suggest 
```suggestion
    const [contentTchapAccessRule, setTchapAccessRule] = useLocalEcho<IAccessRuleEventContent>(
```

odelcroi:
I guess that we need heavy testing on this PR to approve it or would end to end testing be easy to implement?

mcalinghee:
> I guess that we need heavy testing on this PR to approve it or would end to end testing be easy to implement?

[EDIT] As discussed ,we will look at this later on , we need to check unit test

src/lib/createTchapRoom.ts
```diff
@@ -23,41 +19,61 @@ export default class TchapCreateRoom {
...
```
estellecomment, line 40:
```suggestion
                // Here we could have used createRoomOpts.accessRule directly, but since accessRules are a custom Tchap event, it is ignored by later code. So we use createRoomOpts.initial_state, which works properly.
                createRoomOpts.initial_state.push({
```

estellecomment:
I cannot invite externs. I get a silent fail, nothing happens. 

Running locally, with prod backend, after a clearing of node_modules and rebuild : 
Error in console : `t is undefined`. Happens here : https://github.com/matrix-org/matrix-react-sdk/blob/68b3fd78c2759b6b1c6de6f86c93a3e021db36ec/src/components/views/dialogs/InviteDialog.tsx#L929-L930

Reproduce : create a room (with or without externs). Invite an extern by email.
Same thing happens whether the extern account exists already or not.

Edit : I will try testing on the review app tomorrow, in case it is a local problem. (will move review app to prod backend for testing)

estellecomment:
Missing translation in french : 
image

src/components/views/settings/TchapJoinRuleSettings.tsx
mcalinghee: cela permet d'assigner `accessRule` à `undefined` si `contentAccessRule` est lui aussi `undefined

Si on enleve `{}`, on peut avoir l'erreur suivante : `Error: Cannot read properties of undefined (reading 'rule')`

mcalinghee:
> Missing translation in french : image

Translation was added

estellecomment:
(closing and reopening PR to trigger the review app, it had been deleted from lack of activity)

estellecomment:
When testing with my account, on prod backend, it works when you paste the email, rather than type it.
Works with externs who already have an account, or not.

There are issues, unrelated to this PR, with :
 - when you type the email and press enter or space, the invite dialog crashes.
 - inviting fails sometimes because of a 403 on GET https://matrix.agent.dinum.tchap.gouv.fr/_matrix/identity/v2/terms. This is the client trying to get the user to approve the terms and conditions of the identity server. This should not be happening, terms in Tchap do not need to be approved. Seen on a user created in v2 but only used in v4.
 
Will file separate issues.

So, testing passes !

estellecomment:
Other issue : 

 - in a room without externs, the toggle appears disabled, when actually it does work.
image

 - when the toggle is disabled, there is no explanation about why.
 
 I'll look into them now.

estellecomment:
About the disabled toggle : it's actually fine. 
It's just that the color for a toggle that's enabled and off is grey (which I associate with disabled). It's consistent with the rest of the app.

estellecomment:
For the extra info, it's complicated to add because the toggle is part of the "Public" radiobutton description.
We should move it out of the "Public" radiobutton. 
Not doing in this PR.

PR #176, opened 2022-09-15
E2E tests in CI

estellecomment:
E2E tests in CI !!

Login test is failing, not sure why, but it's happening locally as well, so it's out of scope of this PR.
(Maybe because of the test user, not set up properly ? looking into in for the next PR)

### DIFF
--- a/.github/workflows/cypress.yaml
+++ b/.github/workflows/cypress.yaml
@@ -6,8 +6,8 @@ on:
   pull_request: { }
   push:
     branches: [ develop_tchap, master_tchap ]
-  repository_dispatch: # :TCHAP: do we need this ?
-    types: [ element-web-notify ]
+  #repository_dispatch: # :TCHAP: we don't use this
+  #  types: [ element-web-notify ]
 env:
   # :TCHAP: not needed for fetchdep.with.version.sh but leaving them otherwise it breaks.
   # These must be set for fetchdep.sh to get the right branch
--- a/.github/workflows/cypress.yaml
+++ b/.github/workflows/cypress.yaml
@@ -6,8 +6,6 @@ on:
   pull_request: { }
   push:
     branches: [ develop_tchap, master_tchap ]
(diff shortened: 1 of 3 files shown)


.github/workflows/cypress.yaml
```diff
@@ -0,0 +1,54 @@
+# Run the tests
+# Todo : reuse a build from another action, instead of rebuilding for nothing.
+# tchap modification : use "./scripts/layered.with.version.sh" instead of "./scripts/layered.sh"
+name: Cypress End to End Tests
+on:
+  pull_request: { }
+  push:
+    branches: [ develop, master ]
```
mcalinghee: est ce qu'il faudrait rajouter `develop_tchap`?

mcalinghee:
C'est vraiment trop cool!

.github/workflows/cypress.yaml
estellecomment: Removed developed and master, which are element branches, and added develop_tchap and master_tchap, which we are using. (master_tchap actually doesn't exist yet since we haven't released yet)

We could even block commits to master and develop, to be sure things are clear...

mcalinghee: Yes we could but that would be in a different workflow I guess

.github/workflows/cypress.yaml
```diff
@@ -0,0 +1,54 @@
...
+  repository_dispatch: # :TCHAP: do we need this ?
```
mcalinghee: We could remove this as I don't think we will use this yet.
I think it was used by an external CI to trigger the workflow.

More informations on dispatch : 
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#repository_dispatch

estellecomment:
Tests are broken, but they run in CI. So I'm merging.

estellecomment:
Tests are broken, but they run in CI. So I'm merging.

PR #191, opened 2022-09-21
add create release process

mcalinghee:


mcalinghee:
See: https://github.com/tchapgouv/tchap-web-v4/wiki#release-process

PR #199, opened 2022-09-23
New login - e2e test

estellecomment:
## Checklist

* [ ] Tests written for new code (and old code if feasible)
* [ ] Linter and other CI checks pass
* [ ] Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/vector-im/element-web/blob/develop/CONTRIBUTING.md))

### DIFF
--- a/config.sample.json
+++ b/config.sample.json
@@ -1,23 +1,28 @@
 {
     "default_server_config": {
         "m.homeserver": {
-            "base_url": "https://matrix.i.tchap.gouv.fr",
-            "server_name": "Internes - Preprod"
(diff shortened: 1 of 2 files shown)


estellecomment:
"AssertionError: Timed out retrying after 15000ms: Expected to find element: `#mx_LoginForm_email`, but never found it."
Not sure why, maybe because the initial page load is too slow. 
The build (yarn start) is also very slow, sometimes we get a timeout after 90s and crash before running any tests. 
Looks like the CI machines are a bit weak... Probably increase the timeout for the whole setup, and in particular for the first load.

PR #208, opened 2022-09-23
In logout dialog, hide secure storage option

estellecomment:
I wanted to put it in the same patch as hide-secure-storage, to have everything together, but I couldn't find a regex to select the right files for patch-package. So I decided to keep it simple and make a separate patch with similar name.

Not tested yet. I have to figure out how to make this dialog appear.
On a "clean" account without cross-signing, it doesn't appear, whether you have multiple sessions or single session.


odelcroi:
got this panel, seems good to me : 
Capture d’écran 2022-09-26 à 10 08 08

PR #221, opened 2022-09-26
Fix login problems

### LINKED ISSUE #201: On failed login, no error message displays
With my cross3 account, on localhost, prod backend.
Console says - `Login failed M_FORBIDDEN: Unknown message`
Server request returns 403, no error message.

Happens when account does not exist. 
When email exists and password is wrong, the error message displays alright.

(not done investigating, this is preliminary findings)

estellecomment:
Fixes a concurrency problem (initLoginLogic is actually async, so add a few awaits where appropriate, otherwise it continues to run in the background and makes mischief). This explains why error messages where not displaying : the component was immediately rerendered without an error.

Also added a user-friendly error message for TOO_MANY_REQUESTS, since it happens after only 5 requests, it's likely to be seen by users.

image

Fixes https://github.com/tchapgouv/tchap-web-v4/issues/201
(we can reopen if new problems appear)


src/i18n/strings/tchap_translations.json
```diff
@@ -88,5 +88,9 @@
   "We have compiled a list of the questions our users ask the most frequently. Your question may be answered there.": {
     "en": "We have compiled a list of the questions our users ask the most frequently. Your question may be answered there.",
     "fr": "Nous avons constitué une liste des question les plus fréquentes posées par nos utilisateurs. Votre question a peut-être déjà une solution."
+  },
+  "You have made to many requests and have been blocked. Try again later.": {
+    "en": "You have made to many requests and have been blocked. Try again later.",
+    "fr": "Vous avez fait trop de requêtes et avez été bloqué. Réessayez plus tard."
```
odelcroi: scary!  "requetes"->"tentatives" , "bloqué temporairement" ?

mcalinghee: Nous avons ceci dans la v2 : 
`Your last three login attempts have failed. Please try again in 30 minutes.`

estellecomment: after testing, it's 3 attempts, and the delay for unblocking is variable, between 5 and 8 minutes. It probably depends  how many attempts you made recently ? 
So I used "Your last three login attempts have failed. Please try again in a few minutes."
Thanks for suggestions ! 👍

estellecomment:
Figuring out a process to merge patches. Do not merge for now.

estellecomment:
My confusion was because part of the concurrency fix had already been merged as part of fix/better-forgot-password, which was reusing the same functions.

estellecomment:
My confusion was because part of the concurrency fix had already been merged as part of fix/better-forgot-password, which was reusing the same functions.

PR #242, opened 2022-10-03
disable secure storage in userInfo>device panel

### LINKED ISSUE #231: Disable secure storage from UserInfo
### Steps to reproduce

1. Click on a Room where you belong
2. Click on participants>your profile
3. Click on your devices
4. Click on a device
5. Secure Storage process launches

Capture d’écran 2022-10-03 à 12 15 54

#### What did you expect?
- if device is verified, nothing should happen
- if device is not verified, a local verification should start 

#### What happened instead?
- secure storage process started

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Application version

_No response_

### Homeserver

_No response_

### Will you send logs?

No

odelcroi:
- the setting e2ee.manuallyVerifyAllSessions is hidden in user Panel
- overriding the setting e2ee.manuallyVerifyAllSessions to 'true' will make secure storage setup inactive when clicking into a device in userInfo

Before : 
Capture d’écran 2022-10-03 à 12 15 54

After : 

Capture d’écran 2022-10-03 à 12 22 01


patches/disable-secure-storage/matrix-react-sdk+3.57.0.patch
```diff
@@ -0,0 +1,22 @@
...
+ import { OverwriteLoginPayload } from "./dispatcher/payloads/OverwriteLoginPayload";
++import TchapUserSettings from '../../../src/util/TchapUserSettings';
+ 
+ const HOMESERVER_URL_KEY = "mx_hs_url";
+ const ID_SERVER_URL_KEY = "mx_is_url";
+@@ -819,6 +820,9 @@ async function startMatrixClient(startSyncing = true): Promise<void> {
+         // before the client.
+         await EventIndexPeg.init();
+         await MatrixClientPeg.start();
++        //:tchap: override user settings after the client init
```
jdauphant: Is that possible to listen to the dispatcher instead of using a patch ?
I see 2 events sent in this function : ` dis.dispatch({ action: 'will_start_client' }, true);` or `dis.dispatch({ action: 'client_started' });`

jdauphant (commented):
I'm not able to see if "e2ee.manuallyVerifyAllSessions" is the good way to do it.
The settings is local or stored on the server user profil ?

patches/disable-secure-storage/matrix-react-sdk+3.57.0.patch
odelcroi: good point, I'll give it a try

odelcroi:
e2ee.manuallyVerifyAllSessions is used and stored at device level, so overriding it affects only the current device

jdauphant:
This will cause trouble when we will activate cross-signing in the following months ?

jdauphant (approved):

PR #247, opened 2022-10-13
Make globe icon appears only on Forum

odelcroi:
Right now, globe icon is shown on forum and private room regardless of encryption or not. 

Before : 
Capture d’écran 2022-10-13 - v2

After,  globe is only shown on not encrypted room = FORUM on Tchap
Capture d’écran 2022-10-13 à 10 46 29

### DIFF
--- a/patches/patches.json
+++ b/patches/patches.json
@@ -1,4 +1,10 @@
 {
+  "show-icon-on-public-room": {
+    "package": "matrix-react-sdk",
+    "files": [
+      "src/components/views/avatars/DecoratedRoomAvatar.tsx"
+    ]
+  },
   "rename-exported-keys": {
     "package": "matrix-react-sdk",
     "files": [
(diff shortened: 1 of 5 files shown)


mcalinghee (commented):
**Test :**
_Before Change_
1. In v2, create a private room
2. Activate link access to this room
3. In v4, we can see the room with the globe icon
image

_After Change_
1. In v2, create a private room
2. Activate link access to this room
3. In v4 with the new change, we can see the room without the globe icon
image

_No Regression_
- the Forum/Public Room has the globe icon
- Private room does not have the globe icon
- Private room with externals does not have the globe icon

image

=> Result : OK

mcalinghee:
**Test :**
_Before Change_
1. In v2, create a private room
2. Activate link access to this room
3. In v4, we can see the room with the globe icon
image

_After Change_
1. In v2, create a private room
2. Activate link access to this room
3. In v4 with the new change, we can see the room without the globe icon
image

mcalinghee (approved):

Issue #186, opened 2022-09-19
Roomlist displays private and public rooms in the same way


estellecomment:
This is confusing for users. Both forums and private rooms are in the same section of roomlist, and with the same icon.

odelcroi:
+ we do not see at a glance if a room allows external users

odelcroi:
@areox-net @Caroline-lawson  Could we work on some UI components to enhance UX on this topic? Should we take the icon from V2 version?

Caroline-lawson:
Another linked problem is that people I spoke with did not know that forums are not crypted conversations. 

Remind me what the icon was in V2?

If it's the padlock, then it makes sense if it's on all types of group conversations. The problem is that the colours are counter intuitive : 
- red for private rooms: they are encrypted and therefore safe in the context of Tchap - a green closed padlock should make more sense here?
- orange for private rooms open to external users: they are also encrypted - orange closed padlock?
- green for forums: they are open to all Tchap users and not encrypted - red open padlock?

If it isn't the padlock - please ignore.

The "wave" icon didn't make much sense to the people I spoke to.

Open to other suggestions of course.

areox-net:
That's a recurring theme. Tchap v1 highlighted this difference by design, with avatar shapes, icons and - later - color labels for redundancy. We moved to a simpler roomlist as we moved toward Element.
After all these iterations, it is my opinion that the best solution is to separate Forums from other conversations. They are something different both in security level and use.
In this specific case, the web UI is particularly suited for this solution, since it already has sections. 
Is it technically difficult to add a "Forums" section? 

The problem still remains with external users. However, last time we talked about this it was decided that this information is relevant once we are inside a room rather than in the roomlist. So my proposal was to display this information in the room header, both on mobile and web.

odelcroi:
> Is it technically difficult to add a "Forums" section?

I would say it is not straightfoward as we need to implement some logic in the right panel. I will investigate it, but in the mean time, could we propose a quick-win solution to be deployed in the coming days?

> So my proposal was to display this information in the room header, both on mobile and web.
Sure, could you propose a design to do so?

areox-net:
@odelcroi A quick solution for now would be to use the same icon that we use on mobile for Forums 
![Screenshot 2022-10-12 at 11 41 40](https://user-images.githubusercontent.com/1255251/195308947-3cb2bcfd-a334-4c09-940b-8718cd83b961.png)

For the header, my old proposals looked like this 
[...]

PR #255, opened 2022-10-18
Change icon to green unlocked for public rooms

estellecomment:
#186 

image


estellecomment:
The size of the lock icon looks a bit small to me, but it was difficult to make it bigger : there is a white circle below the lock icon, which is harder to remove. So the size of the lock icon has to fit within this circle. 
We can do better later if needed, I went for the easy win.

estellecomment:
While I'm at it, I'm changing the green icon in the room creation dialog as well : 
image

cc @areox-net

odelcroi (approved):
Nice!

PR #256, opened 2022-10-18
add room header when externals are allowed

### LINKED ISSUE #186: Roomlist displays private and public rooms in the same way

odelcroi:
With the header : 
Capture d’écran 2022-10-18 à 16 30 24

Test : 
- [x] 1 : create a room with external allowed, check that the header is displayed
- [x] 2 : create a private room without external allowed, check that the header is NOT displayed
- [x] 3 : create a private room, after the creation, allow external in the settings, check that the header is displayed

e2e : 
- [x] 1 : create a room with external allowed, check that the header is displayed
- [x] 2 : create a private room without external allowed, check that the header is NOT displayed

UI : 
- [ ] : check responsiveness -> N/A


src/components/views/rooms/TchapExternalRoomHeader.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { _t } from "matrix-react-sdk/src/languageHandler";
+
+import { TchapRoomAccessRule, TchapRoomAccessRulesEventId } from '../../../@types/tchap';
+import TchapRoomUtils from '../../../util/TchapRoomUtils';
+
+export interface IProps {
+    room?: Room;
+}
+
+export function checkExternalAllowed(room: Room): boolean {
```
estellecomment, line 13: je mettrais bien cette fonction dans TchapRoomUtils

src/components/views/rooms/TchapExternalRoomHeader.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+    //listen to events to refresh the value if needed
+    useTypedEventEmitter(room.currentState, RoomStateEvent.Events, (ev: MatrixEvent) => {
+        console.log(ev.getType());
+        if (ev.getType() === TchapRoomAccessRulesEventId) {
+            setExternalAllowed(checkExternalAllowed(room));
+        }
+    });
+    //refresh the state after rendering
+    useEffect(() => {
```
estellecomment: Hmmm.
I think the react lifecycle is : state change -> rendering -> effect.
Here the effect contains a state change (setExternalAllowed). Does that make an infinite loop ? state change -> rendering -> effect -> state change -> ...

I think that the rendering already uses isExternalAllowed to render the appropriate UI, so the effect is not necessary.

estellecomment: I tried removing it, the status is still updated when you change externs status !

estellecomment:
cypress is broken locally and in CI, but I can't figure out why. Later.

src/components/views/rooms/TchapExternalRoomHeader.tsx
odelcroi: si tu veux :)

src/components/views/rooms/TchapExternalRoomHeader.tsx
odelcroi: nice

odelcroi: > Does that make an infinite loop ? state change -> rendering -> effect -> state change -> ...
the effect is trigger only if room is updated  as mentioned in the comment

odelcroi: ok to remove it, it works well without it :)

PR #258, opened 2022-10-18
Display icons in roomlist for the 3 room types

### LINKED ISSUE #186: Roomlist displays private and public rooms in the same way

estellecomment:
Issue : https://github.com/tchapgouv/tchap-web-v4/issues/186

Screen Shot 2022-10-18 at 5 04 57 PM

Tooltips when hovering on the icon : 

image

image

image


estellecomment:
@areox-net Je continue de te tenir au courant, crie si quelque chose ne va pas :) 

Review app (backend de dev, il faut un compte de dev, pas le compte de prod) : https://tchap-web-staging-v4-pr258.osc-secnum-fr1.scalingo.io/

estellecomment:
Our icons are not accessible for color-blind people : two of them have the same color.

areox-net:
> Our icons are not accessible for color-blind people : two of them have the same color.

That's true. However, they are used together with the related labels. Actually, today the "New Room" screen is the only place where they are displayed. So I don't think this is a bug issue.

mcalinghee (approved):
Tested and approved

PR #269, opened 2022-10-21
Upgrade/element web v1.11.10

estellecomment:
Interesting fixes : 
Notification fail (may explain our notification problem) : https://github.com/vector-im/element-web/issues/23456

PR #271, opened 2022-10-21
Remove location button if map server is not configured

estellecomment:
This changed could be pushed in element-web directly, probably.

We will need to remove Voice Message button soon, which will need a patch in MessageComposer as well. Watch out for patch clash ! (will probably just put the two in a single patch)

(Note that the patch is made against react-sdk 3.58.1, which is not in develop_tchap yet, wait for merge of element-web upgrade https://github.com/tchapgouv/tchap-web-v4/pull/269)

### DIFF
--- a/patches/hide-location-if-no-map-server/matrix-react-sdk+3.58.1.patch
+++ b/patches/hide-location-if-no-map-server/matrix-react-sdk+3.58.1.patch
@@ -1,38 +1,21 @@
 diff --git a/node_modules/matrix-react-sdk/src/components/views/rooms/MessageComposer.tsx b/node_modules/matrix-react-sdk/src/components/views/rooms/MessageComposer.tsx
-index cf0fe3f..c55d062 100644
+index cf0fe3f..41a3bc0 100644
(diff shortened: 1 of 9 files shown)


mcalinghee (approved):

mcalinghee (approved):

patches/hide-location-if-no-map-server/matrix-react-sdk+3.58.1.patch
```diff
@@ -0,0 +1,38 @@
...
+         const showSendButton = !this.state.isComposerEmpty || this.state.haveRecording;
++        // :TCHAP: do not show location button if map server is not configured.
++        let isMapConfigured;
++        try {
++            findMapStyleUrl();
++            isMapConfigured = true;
++        } catch (e) {
++            isMapConfigured = false;
++        }
++        const showLocationButton = !window.electron && isMapConfigured;
```
odelcroi: should this logic be placed in a util class?

estellecomment: yes ! done

Issue #275, opened 2022-10-24
Implement registration


estellecomment:
#### What would you like to do?
Create da account man

#### Why would you like to do it?
Wanna do the tchap thang

#### How would you like to achieve it?
add the call to find the user's homeserver, and fit it in somewhere in the flow

### Have you considered any alternatives?

reuse v2 registration : https://github.com/tchapgouv/tchap-web-v4/issues/152

PR #276, opened 2022-10-24
Implement registration

estellecomment:
Fixes https://github.com/tchapgouv/tchap-web-v4/issues/275

The full flow in element-web is a bit better than the flow in v2 : when the user clicks the link in email, they are redirected to the login page.

However, this implementation is reverting back to the v2 implementation, where the clicked link displays a confirmation page served by the backend, and then the user has to go back to login by hand.

This is because of a bug : after the redirect, element-web gets the status of the account creation by doing a `POST /register`. But in the Tchap case this request returns an error, `M_THREEPID_IN_USE: email is already in use`, even though the account is successfully created. https://github.com/tchapgouv/tchap-web-v4/issues/280
So we work around this problem for the moment by removing the redirect, and thus we are back to the v2 flow for now.

Review notes : 
 - For easier reviewing of the patch, you can look at the PR on react-sdk instead : https://github.com/tchapgouv/matrix-react-sdk-fork/pull/2/files
 - the review app is on the prod backend. Mainlined backends fail : https://github.com/tchapgouv/tchap-web-v4/issues/281

### DIFF
--- a/config.dev.json
+++ b/config.dev.json
@@ -56,7 +56,7 @@
         "breadcrumbs": true,
         "UIFeature.voip": false,
         "UIFeature.shareSocial": false,
-        "UIFeature.registration": false,
+        "UIFeature.registration": true,
         "UIFeature.urlPreviews": false,
         "UIFeature.widgets": false,
         "UIFeature.shareQrCode": false,
--- a/config.preprod.json
+++ b/config.preprod.json
@@ -46,7 +46,7 @@
         "breadcrumbs": true,
         "UIFeature.voip": false,
         "UIFeature.shareSocial": false,
-        "UIFeature.registration": false,
+        "UIFeature.registration": true,
         "UIFeature.urlPreviews": false,
         "UIFeature.widgets": false,
         "UIFeature.shareQrCode": false,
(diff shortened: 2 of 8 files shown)


odelcroi:
Test plan : (tests done by Estelle on review app on prod backend)

- [x] 1) create account with allowed domain

❌ create account with non allowed domain -> email is not sent, but :
  - error message is in english ("Your email is not authorized to register on this server")
  - "email has been sent" is briefly displayed before error message, so it's confusing
  Filed https://github.com/tchapgouv/tchap-web-v4/issues/288
- [x] create account with allowed domain but not existing email -> "email has been sent" and it probably bounces somewhere
- [x] ... with low security password
- [x] 1) + click on register email link
- [x] 1) + click twice on register email -> "La validation de l'email a échouée : le lien a expiré, ou il n'est pas valide."
- [x] 1) + click on resend email : two emails should arrive. Clicking first link should work.
- [x] 1) + click on resend email : two emails should arrive. Clicking second link should work.
- [x] do 1) twice, to generate two emails. Click on either of the emails and it should work.

From the test plan, another interesting use case is : 

- [x] TG_CC_5 : invite external user from tchap to create a new account
  - you have to write the first message to create the room to send the email. Before the first message, the room isn't created and the email isn't sent.
- [x] external user who is not invited cannot create account

estellecomment:
Merging with the existing issue, it's better than nothing and good enough for a beta.

PR #277, opened 2022-10-25
create a tchap-light and tchap-dark theme

### LINKED ISSUE #273: Choice of themes should be just "light" or "dark"
Currently we have "light" "dark" and "tchap", which is weird.

odelcroi:
- build two themes tchap-light and tchap-dark
- make them customizable with css variables and css overrides inside tchap project
- remove legacy tchap theme
- make sure css hot reload does work a bit (anyway for a better experience with themes disable hot reloading in .env file)

TODO : 
- [ ] dark theme is missing the container "mx_BackdropPanel"
- [x] all css vars are not defined properly, because of the logic `function setCustomThemeVars(customTheme: ICustomTheme): void ` -> avatars color is not grey as expected but fallback to default colors

## Light theme
Capture d’écran 2022-10-26 à 17 26 14

## Dark theme
Capture d’écran 2022-10-27 à 11 51 49


package.json
```diff
@@ -52,14 +52,14 @@
...
-    "lint:style": "stylelint \"res/css/**/*.pcss\"",
+    "lint:style": "stylelint \"res/**/*.pcss\"",
     "test": "jest",
     "test:cypress": "cypress run",
     "test:cypress:open": "cypress open",
     "coverage": "yarn test --coverage",
     "analyse:unused-exports": "node ./scripts/analyse_unused_exports.js",
     "postinstall": "./scripts/apply_patches.sh",
-    "patch-package": "patch-package",
+    "patch-package": "patch-package",
```
estellecomment:
```suggestion
    "patch-package": "patch-package",
```

webpack.config.js
```diff
@@ -20,25 +20,14 @@ if (!process.env.VERSION) {
...
 function getActiveThemes() {
-    // Default to `light` theme when the MATRIX_THEMES environment variable is not defined.
-    //const theme = process.env.MATRIX_THEMES ?? 'light';
-    // :TCHAP: We need both themes to avoid "unknown theme light-custom" or "unknown theme light"
-    // being thrown on login page.
-    // Browsers recover, but e2e tests crash, so we add both themes in default here.
-    const theme = process.env.MATRIX_THEMES ?? 'light,light-custom';
-    return theme.split(',').map(x => x.trim()).filter(Boolean);
+    //:tchap: activate only light and dark themes
+    return ["light", "dark"];
```
estellecomment: remove MATRIX_THEMES var from .env.sample ? I think we don't use it anymore.
(and also CSS_HOT_RELOAD if we don't support it any more)

webpack.config.js
```diff
@@ -110,7 +99,11 @@ module.exports = (env, argv) => {
     const ACTIVE_THEMES = getActiveThemes();
     function getThemesImports() {
         const imports = ACTIVE_THEMES.map((t) => {
-            return cssThemes[`theme-${ t }`].replace('./node_modules/', ''); // theme import path
+            if (useHMR) {
```
estellecomment, line 114: Why do we add useHMR here ?

webpack.config.js
```diff
@@ -20,25 +20,14 @@ if (!process.env.VERSION) {
...
-    // CSS themes
-    "theme-legacy-light": "./node_modules/matrix-react-sdk/res/themes/legacy-light/css/legacy-light.pcss",
-    "theme-legacy-dark": "./node_modules/matrix-react-sdk/res/themes/legacy-dark/css/legacy-dark.pcss",
-    "theme-light": "./node_modules/matrix-react-sdk/res/themes/light/css/light.pcss",
-    "theme-light-high-contrast":
-        "./node_modules/matrix-react-sdk/res/themes/light-high-contrast/css/light-high-contrast.pcss",
-    "theme-dark": "./node_modules/matrix-react-sdk/res/themes/dark/css/dark.pcss",
-    "theme-light-custom": "./node_modules/matrix-react-sdk/res/themes/light-custom/css/light-custom.pcss",
-    "theme-dark-custom": "./node_modules/matrix-react-sdk/res/themes/dark-custom/css/dark-custom.pcss",
+    //:tchap: use path from our project
```
estellecomment, line 31: here and below, I think it's better to comment the code we don't use than to remove it. It looks ugly but the future upgrades will be easier (if you remove it, it might re-appear in a future merge, by mistake, we've seen it happen)

estellecomment:
Testing
 - [x]  visual inspection of light page
 - [x] visual inspection of dark page
 - [ ] are e2e tests broken, on login page especially ?
 
---

When CSS_HOT_RELOAD=1, we have a partly-themed login page. It's even more obvious when you have dark theme :)) But it was already partly themed before this PR, there were both blue and green buttons.
When CSS_HOT_RELOAD=0 it's fine. Prod has CSS_HOT_RELOAD=0 I suppose.
image

---
With CSS_HOT_RELOAD=1, the theme changing doesn't work. We end up stuck in light theme.
Don't know if it's a big issue or not.

estellecomment:
Font has changed, but that's the same as element-web, it's fine. We can change the font for Tchap as part of theme customisation.

webpack.config.js
odelcroi: there is a hack when using hot reloading (useHMR) where the css are imported from another file (path is taken from src/vector/devcss.ts), that's why we need to override this path

webpack.config.js
odelcroi: goodpoint

webpack.config.js
odelcroi: CSS_HOT_RELOAD works but sometimes we do need to restart the app

estellecomment:
Maybe a problem with buttons still green in the dark theme, not sure, but still it's a progress, so merging.

jdauphant:
@areox-net will be happy of this PR ;)

jdauphant:
Close #277

PR #284, opened 2022-10-28
Customise rageshake reports to add user email

estellecomment:
- prefix rageshake reports in crisp with [tchap-web] instead of [element-web]
- add user email in the report

Would be nice to see if the rageshake server, which makes the report email out of the data, can add the user email as a reply-to field, so that crisp gets the reply email automatically. Out of scope here.


estellecomment:
ah zut j'ai pété le version number

estellecomment:
ah non c'est bon

estellecomment:
ah non c'est bon

odelcroi:
Could you tested it successfully ? (I couldn't, ticket was not created in crisp) If so, looks good to me

aulamber (approved):

estellecomment:
> Could you tested it successfully ? (I couldn't, ticket was not created in crisp) If so, looks good to me

It only works with prod config (the review app has dev config). But yes, I did make it work. Merging !

PR #287, opened 2022-11-03
add a handler to refresh client store when migrating from v2

### LINKED ISSUE #278: My contacts can not uncrypt my messages
### Steps to reproduce

In some cases, after migrating to v4, some users reported that their contacts could not uncrypt their messages. 

Clearing the cache, restarting the client, hard refreshing the page would solve the problem. 

Investigations showed that their keys were not sent to their contacts, which was solved by restarting the client (either clearing the cache or refreshing the page)

#### What did you expect?
- "keys" should always sent to my contacts to uncrypt my messages

#### What happened instead?
- they weren't

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Application version

_No response_

### Homeserver

_No response_

### Will you send logs?

No

odelcroi:
- At application loading, the sync store can cleared if needed.

- When upgrading from v2 to v4, the sync store will be cleared and the page will reload automatically to force a new  synchronisation.  

- the application version is stored in localStorage at key : tchap_app_version

Capture d’écran 2022-11-03 à 14 42 32

### DIFF
--- a/src/app/initTchap.ts
+++ b/src/app/initTchap.ts
@@ -1,0 +1,53 @@
+import defaultDispatcher from "matrix-react-sdk/src/dispatcher/dispatcher";
+import { ActionPayload } from "matrix-react-sdk/src/dispatcher/payloads";
+import PlatformPeg from "matrix-react-sdk/src/PlatformPeg";
+
+import TchapClientUtils from "../util/TchapClientUtils";
+import TchapUserSettings from "../util/TchapUserSettings";
+
+/**
+ * Force refresh after the client has started
+ */
+export function attachRefreshHandler() {
+    const clearCacheAndReloadId = defaultDispatcher.register(
+        (payload: ActionPayload) => {
+            if (payload.action === "client_started") {
+                //unregister callback once the work is done
+                defaultDispatcher.unregister(clearCacheAndReloadId);
+                //:tchap: use localstorage instead of matric idDB ?
+                TchapClientUtils.clearCacheAndReload();
+            }
+        },
+    );
+}
+
+/**
+ * Save app version to localstorage after the client has started
+ */
+export function attachVersionAppHandler() {
+    const saveVersionId = defaultDispatcher.register(
+        (payload: ActionPayload) => {
+            if (payload.action === "client_started") {
+                //override user settings
+                TchapClientUtils.saveAppVersion(PlatformPeg.get());
+                //unregister callback once the work is done
+                defaultDispatcher.unregister(saveVersionId);
+            }
+        },
+    );
+}
+/**
+ * Override user settings after the client has started
+ */
+export function attachOverideUsersOptionHandler() {
+    const registerId = defaultDispatcher.register(
+        (payload: ActionPayload) => {
+            if (payload.action === "client_started") {
+                //override user settings
+                TchapUserSettings.override();
+                //unregister callback once the work is done
+                defaultDispatcher.unregister(registerId);
+            }
+        },
+    );
+}
--- a/src/app/initTchap.ts
+++ b/src/app/initTchap.ts
@@ -2,35 +2,48 @@ import defaultDispatcher from "matrix-react-sdk/src/dispatcher/dispatcher";
 import { ActionPayload } from "matrix-react-sdk/src/dispatcher/payloads";
 import PlatformPeg from "matrix-react-sdk/src/PlatformPeg";
 
-import TchapVersionManagement from "../util/TchapClientUtils";
+import TchapVersionManagement from "../util/TchapVersionManagement";
 import TchapUserSettings from "../util/TchapUserSettings";
+import TchapUIFeature from "../util/TchapUIFeature";
 
 /**
- * Determine weither the app needs a refresh after loading
- * @param indexedDB the indexDb interface
+ * Determine weither the app needs a clearCacheAndReload after loading. We do it when upgrading from v2 to v4, to avoid weird keys bugs.
  * @returns Promise(true) if a refresh is needed, Promise(false) in other cases
  */
-export async function needsRefreshForVersion4(indexedDB: IDBFactory): Promise<boolean> {
-    if (!indexedDB) {
-        Promise.reject("indexDb is undefined");
-    }
-
-    //if tchap version is unknown, refresh
-    if (!TchapVersionManagement.getAppVersion() === null) {
-        return Promise.resolve(true);
-    }
-
-    //read version of tchap app in localstorage, if it does start with 4, don't refresh
-    if (TchapVersionManagement.getAppVersion() && TchapVersionManagement.getAppVersion().startsWith("4")) {
+export async function needsRefreshForVersion4(): Promise<boolean> {
+    //check if the feature is activated
+    if (!TchapUIFeature.activateClearCacheAndReloadAtVersion4) {
         return Promise.resolve(false);
     }
 
+    let indexedDB;
     try {
-        //read version of matrix-js-sdk:riot-web-sync, if it is lower than version 4, refresh is needed
-        const version = await TchapVersionManagement.getStoreVersion(indexedDB, TchapVersionManagement.SYNC_STORE_NAME);
-        return Promise.resolve(version < 4);
+        indexedDB = window.indexedDB;
+    } catch (e) {}
+
+    if (!indexedDB) {
+        //not sure why it could happen, do not refresh in this case (safety first, avoid creating an infinite clearCacheAndReload loop !)
+        Promise.resolve(false);
+    }
+
+    // Read version of tchap app in localstorage
+    // if it is strictly less to version 4.x.y, do refresh
+    if (TchapVersionManagement.getAppVersion()) {
+        const previousAppVersion: string = TchapVersionManagement.getAppVersion();
+        const previousAppMajorVersion: number = parseInt(previousAppVersion.charAt(0), 10);
+        return Promise.resolve(!isNaN(previousAppMajorVersion) && previousAppMajorVersion < 4);
+    }
+
+    // If there is no tchap app version, it could be that this is a first install of tchap, or that saving the version has failed in the past, or
+    // that the previous install of tchap was before v4. We only want to refresh when we are upgrading v2->v4.
+    try {
+        // Read version of matrix-js-sdk:riot-web-sync, if it is lower than version 4, refresh is needed.
+        // Note : version 4 of the store happens to coincide with tchap-web v4, but it's unrelated.
+        const previousStoreVersion: number =
+            await TchapVersionManagement.getStoreVersion(indexedDB, TchapVersionManagement.SYNC_STORE_NAME);
+        return Promise.resolve(previousStoreVersion < 4);
     } catch (error) {
-        //if store did not exists or a technical error occured do not refresh (safety first)
+        // Do not refresh (safety first, avoid creating an infinite clearCacheAndReload loop !)
         console.warn(error);
         return Promise.resolve(false);
     }
--- a/src/app/initTchap.ts
+++ b/src/app/initTchap.ts
@@ -2,7 +2,7 @@ import defaultDispatcher from "matrix-react-sdk/src/dispatcher/dispatcher";
 import { ActionPayload } from "matrix-react-sdk/src/dispatcher/payloads";
 import PlatformPeg from "matrix-react-sdk/src/PlatformPeg";
 
-import TchapClientUtils from "../util/TchapClientUtils";
+import TchapVersionManagement from "../util/TchapClientUtils";
 import TchapUserSettings from "../util/TchapUserSettings";
 
 /**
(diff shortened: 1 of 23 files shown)


src/vector/index.ts
```diff
@@ -130,6 +128,15 @@ async function start() {
         await settled(rageshakePromise);
 
         const fragparts = parseQsFromFragment(window.location);
+        //:tchap: get the indexedDB store version
+        let indexedDB;
+        try {
+            indexedDB = window.indexedDB;
+        } catch (e) {}
+
+        const needRefresh = await TchapClientUtils.doesNeedRefresh(indexedDB);
```
weeman1337: This raises an error if `indexedDB` is not set. It will be catched and leads to `Your Element is misconfigured`. Don't know if this is the desired behaviour.

src/app/initTchap.ts
```diff
@@ -0,0 +1,53 @@
...
+                TchapClientUtils.clearCacheAndReload();
+            }
+        },
+    );
+}
+
+/**
+ * Save app version to localstorage after the client has started
+ */
+export function attachVersionAppHandler() {
```
weeman1337: (opinionated) I would merge the handlers, that do not depend on a condition.

src/util/TchapClientUtils.ts
```diff
@@ -0,0 +1,99 @@
...
+                resolve(version);
+            };
+            req.onerror = ev => reject(req.error);
+        });
+    }
+
+    static VERSION_APP_KEY = "tchap_app_version";
+
+    public static async saveAppVersion(platform: BasePlatform) {
+        const localStorage = window.localStorage;
```
weeman1337: In what case can `localStorage` not be defined?

odelcroi: Would it happen if it was not allowed in firefox for instance?

weeman1337: I just ask because then it should be checked at the other place as well https://github.com/tchapgouv/tchap-web-v4/pull/287/files#diff-8582070d01f08e8f48de4b9b84ad2da5f7f704e4e767ee0e8843bb23265cee3aR97

src/vector/index.ts
odelcroi: I've taken the same logic from https://github.com/matrix-org/matrix-react-sdk/blob/f7159b859c9e64d8591c66ea72cf092c7eecb300/src/utils/createMatrixClient.ts#L34

src/util/TchapClientUtils.ts
```diff
@@ -0,0 +1,99 @@
...
+
+export default class TchapClientUtils {
+    static SYNC_STORE_NAME = "matrix-js-sdk:riot-web-sync";
+
+    /**
+     * Determine weither the app needs a refresh after loading
+     * @param indexedDB the indexDb interface
+     * @returns Promise(true) if a refresh is needed, Promise(false) in other cases
+     */
+    public static async doesNeedRefresh(indexedDB: IDBFactory): Promise<boolean> {
```
estellecomment: This function has "4" hardcoded, so I suggest :
```suggestion
    public static async needsRefreshForVersion4(indexedDB: IDBFactory): Promise<boolean> {
```
(another option is to make a customiseable function that we can reuse in the future if this happens again... but I think keeping it simple for now is a good idea to avoid problems.)

src/util/TchapClientUtils.ts
```diff
@@ -0,0 +1,99 @@
+import BasePlatform from "matrix-react-sdk/src/BasePlatform";
+import { MatrixClientPeg } from "matrix-react-sdk/src/MatrixClientPeg";
+import PlatformPeg from "matrix-react-sdk/src/PlatformPeg";
+
+export default class TchapClientUtils {
```
estellecomment:
```suggestion
export default class TchapVersionManagement {
```

src/util/TchapClientUtils.ts
```diff
@@ -0,0 +1,99 @@
...
+     * Determine weither the app needs a refresh after loading
+     * @param indexedDB the indexDb interface
+     * @returns Promise(true) if a refresh is needed, Promise(false) in other cases
+     */
+    public static async doesNeedRefresh(indexedDB: IDBFactory): Promise<boolean> {
+        if (!indexedDB) {
+            Promise.reject("indexDb is undefined");
+        }
+        //read version of tchap app in localstorage, if it does start with 4, don't refresh
+        if (TchapClientUtils.getAppVersion() && TchapClientUtils.getAppVersion().startsWith("4")) {
```
estellecomment: Should it be "4 and above" here ? If version=5, we are over version 4, so no refresh.

src/util/TchapClientUtils.ts
```diff
@@ -0,0 +1,99 @@
...
+            return Promise.resolve(false);
+        }
+
+        try {
+            //read version of matrix-js-sdk:riot-web-sync, if it is older than version 4, refresh is needed
```
estellecomment:
```suggestion
            //read version of matrix-js-sdk:riot-web-sync, if it is lower than version 4, refresh is needed
```

src/util/TchapClientUtils.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    public static async saveAppVersion(platform: BasePlatform) {
+        const localStorage = window.localStorage;
+        if (localStorage) {
+            const version: string = await platform.getAppVersion();
+            localStorage.setItem(TchapClientUtils.VERSION_APP_KEY, version);
+        }
+    }
+
+    public static getAppVersion(): string {
```
estellecomment: here we return a string, whereas getStoreVersion returns a number.
Can we add examples in function doc, so that we know what we are talking about ? "4.0.2_1.11.10" vs. `4` ? 

I would like to be careful, to avoid bugs trying to compare the wrong things and reaching the wrong result.

src/util/TchapClientUtils.ts
estellecomment: can you document why you do this ? In what cases is this necessary ? why do we save tchap_app_version since there is already storeVersion present ?

src/util/TchapClientUtils.ts
```diff
@@ -0,0 +1,99 @@
...
+    * https://github.com/matrix-org/matrix-react-sdk/blob/3c5c2bef6dbac51ce6e1864056523815ca4c38d9/src/components/views/settings/tabs/user/HelpUserSettingsTab.tsx#L308
+     * @returns nothing
+     */
+    public static clearCacheAndReload(): void {
+        console.log(":TCHAP: clearCacheAndReload at client startup");
+
+        if (!PlatformPeg.get()) return;
+
+        MatrixClientPeg.get().stopClient();
+        MatrixClientPeg.get().store.deleteAllData().then(() => {
```
estellecomment: what data does this delete ? 
Is this a problem for users with a single session, who don't have other sessions to get keys from ? (we don't have secure storage in tchap)

(just asking general questions, to make sure we're not doing something crazy)

src/util/TchapClientUtils.ts
```diff
@@ -0,0 +1,99 @@
...
+                // Since we did not provide an explicit version when opening, this event
+                // should only fire if the DB did not exist before at any version.
+                exists = false;
+            };
+            req.onblocked = () => reject(req.error);
+            req.onsuccess = () => {
+                const db = req.result;
+                const version: number = db.version;
+                db.close();
+                if (!exists) {
```
estellecomment: shouldn't we do this deleteDatabase before we get the bd.version ? we are getting the version from a db that doesn't exist, and then ignoring that version, it's hard to understand.

src/app/initTchap.ts
estellecomment: This function attaches the handler, but also does something, and then detaches the handler. So : 
```suggestion
export function queueSaveAppVersion() {
```
or asyncSaveAppVersion, or something like that.

src/app/initTchap.ts
```diff
@@ -0,0 +1,53 @@
...
+                //unregister callback once the work is done
+                defaultDispatcher.unregister(saveVersionId);
+            }
+        },
+    );
+}
+/**
+ * Override user settings after the client has started
+ */
+export function attachOverideUsersOptionHandler() {
```
estellecomment: same remark
```suggestion
export function queueOverideUserSettings() {
```

src/app/initTchap.ts
```diff
@@ -0,0 +1,53 @@
...
+import { ActionPayload } from "matrix-react-sdk/src/dispatcher/payloads";
+import PlatformPeg from "matrix-react-sdk/src/PlatformPeg";
+
+import TchapClientUtils from "../util/TchapClientUtils";
+import TchapUserSettings from "../util/TchapUserSettings";
+
+/**
+ * Force refresh after the client has started
+ */
+export function attachRefreshHandler() {
```
estellecomment: This function attaches the handler, but also does something, and then detaches the handler. So :
```suggestion
export function queueClearCacheAndReload() {
```

src/vector/index.ts
```diff
@@ -130,6 +128,15 @@ async function start() {
         await settled(rageshakePromise);
 
         const fragparts = parseQsFromFragment(window.location);
+        //:tchap: get the indexedDB store version
```
estellecomment: can you document why `needRefresh` check is done here, while the saving of appVersion is done later ?

estellecomment (commented):
wow good investigation, I don't understand eeeeverything but looks cool, let's talk :)

src/util/TchapClientUtils.ts
odelcroi: For the moment, I'll stick to starts with 4 as it is the refresh we need know

src/util/TchapClientUtils.ts
odelcroi: ok

src/util/TchapClientUtils.ts
odelcroi: When opening the db, if it does not exists, it is created, that's why we need to delete it afterward

src/app/initTchap.ts
odelcroi: nice

src/app/initTchap.ts
```diff
@@ -0,0 +1,76 @@
...
+    if (TchapVersionManagement.getAppVersion() && TchapVersionManagement.getAppVersion().startsWith("4")) {
+        return Promise.resolve(false);
+    }
+
+    try {
+        //read version of matrix-js-sdk:riot-web-sync, if it is lower than version 4, refresh is needed
+        const version = await TchapVersionManagement.getStoreVersion(indexedDB, TchapVersionManagement.SYNC_STORE_NAME);
+        return Promise.resolve(version < 4);
+    } catch (error) {
+        //if store did not exists or a technical error occured do not refresh (safety first)
```
estellecomment:
```suggestion
        // if store did not exists or a technical error occured do not refresh (safety first). This happens for example when a clearCacheAndReload has been done : no more store.
```

src/app/initTchap.ts
```diff
@@ -0,0 +1,76 @@
...
+        Promise.reject("indexDb is undefined");
+    }
+
+    //if tchap version is unknown, refresh
+    if (!TchapVersionManagement.getAppVersion() === null) {
+        return Promise.resolve(true);
+    }
+
+    //read version of tchap app in localstorage, if it does start with 4, don't refresh
+    if (TchapVersionManagement.getAppVersion() && TchapVersionManagement.getAppVersion().startsWith("4")) {
```
estellecomment: 4 or more

src/app/initTchap.ts
```diff
@@ -0,0 +1,89 @@
...
+import TchapUIFeature from "../util/TchapUIFeature";
+
+/**
+ * Determine weither the app needs a clearCacheAndReload after loading. We do it when upgrading from v2 to v4, to avoid weird keys bugs.
+ * @returns Promise(true) if a refresh is needed, Promise(false) in other cases
+ */
+export async function needsRefreshForVersion4(): Promise<boolean> {
+    //check if the feature is activated
+    if (!TchapUIFeature.activateClearCacheAndReloadAtVersion4) {
+        return Promise.resolve(false);
```
weeman1337: You can just `return false` here because it is an `async` function. Reduces visual clutter.

src/vector/index.ts
estellecomment: fixed in latest commits, thanks !

src/app/initTchap.ts
estellecomment: done, thanks !

odelcroi:
Testing in dev environment (firefox, chrome, edge)

1) migrate from v2
- from a fresh browser connect to https://www.tchap.incubateur.net, it should propose v2
- login with a user, write any number of encrypted messages to any number of rooms
- hit https://tchap.incubateur.net/join-beta, it should switch to v4
- a automatic reload should happen, check in the logs for `:TCHAP: queue a hard clear cache and reload for this version? true`

2) upgrade from a previous v4
- you should already be connected to a previous version of v4
- hit https://tchap.incubateur.net/
- it should load the new version, without a refresh, check in the logs for `:TCHAP: queue a hard clear cache and reload for this version? false`

3) already upgraded
- you should already be connected to this version of v4
- hit https://tchap.incubateur.net/
- it should load the app, without a refresh, check in the logs for `:TCHAP: queue a hard clear cache and reload for this version? false`

4) connect for the first time to this version
- from a fresh browser connect to https://tchap.incubateur.net/join-beta, it should propose this version of v4
- it should load the app, without a refresh, check in the logs for `:TCHAP: queue a hard clear cache and reload for this version? false`

Firefox/win10
- [x] 1
- [x] 2
- [x] 3
- [x] 4

Edge/win10
- [x] 1
- [x] 2
- [x] 3
- [x] 4

Chrome/win10
- [x] 1
- [x] 2
- [x] 3
- [x] 4

jdauphant:
@odelcroi I just see the implementation.
We should refresh for all from the previous version that are v2 and v4.
We have ~1000 persons that have already migrated to v2 .

So "2." should have been true to refresh all this persons.

Issue #215, opened 2022-09-26
As an external, hide all features I should access


mcalinghee:
### Your use case

As external, I can't :
- Invite
- Create Room/Space
- Be an admin
- Join a public room
- [edit Estelle] modify room settings

### Additional context

Step to implement : 

- Detect if user is externe here : (example here : https://github.com/vector-im/element-web/blob/develop/docs/customisations.md#component-visibility-customisation )
- Hide all UIComponent listed here : https://github.com/matrix-org/matrix-react-sdk/blob/30460943b22840136ddf3bddc8ff2917fc33d536/src/settings/UIFeature.ts#L36
- Implement more UIComponent if needed in Element Web

Inspiration on react-sdk-v4 to determine if an user is external : https://github.com/tchapgouv/matrix-react-sdk-v3/blob/develop/src/tchap/Tchap.ts#L146-L191

More Information on this issue here :
https://github.com/tchapgouv/tchap-web-v4/issues/91

PR #359, opened 2023-01-05
Hide feature InviteUsers to external people #215

jdauphant:
Part of #215

Steps :
- [x] Basic implementation
- [ ] Tested

### DIFF
--- a/src/customisations/TchapComponentVisibility.ts
+++ b/src/customisations/TchapComponentVisibility.ts
@@ -11,6 +11,7 @@ File copied from ComponentVisibility.ts in matrix-react-sdk v3.46.0
 // Populate this class with the details of your customisations when copying it.
 
 import { UIComponent } from "matrix-react-sdk/src/settings/UIFeature";
+import { MatrixClientPeg } from "matrix-react-sdk/src/MatrixClientPeg";
 
 /**
  * Determines whether or not the active MatrixClient user should be able to use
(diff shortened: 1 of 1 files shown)


jdauphant:
This PR isn't a priority, it was made to solve a problem that have been solve in the backend instead.

src/customisations/TchapComponentVisibility.ts
```diff
@@ -23,8 +24,12 @@ import { UIComponent } from "matrix-react-sdk/src/settings/UIFeature";
  * otherwise.
  */
 function shouldShowComponent(component: UIComponent): boolean {
+    const userId = MatrixClientPeg.get().getUserId();
+    const isExternal = userId.endsWith("agent.externe.tchap.gouv.fr") || userId.endsWith("agent.e.tchap.gouv.fr");
```
estellecomment, line 28: It would be nice to have a function "isUserExternal" in the tchap files somewhere, so that we can reuse it later (we have not started extern customisation yet, but there will be many to do)

I would also add dev env, to avoid losing time with this when testing.

I don't know if there is a backend call we can do, or a property in MatrixClientPeg we can check, to know if a user is extern ? If we don't have anything clean like that, then this is fine.

odelcroi: you mean to add a config var to identify which are the external server? I agree it is the same problematic than https://github.com/tchapgouv/tchap-web-v4/pull/59

jdauphant:
I will not go further in this PR.
We can planifiate it in a sprint

odelcroi:
Agreed, closing for now, feel free to reopen

PR #377, opened 2023-01-16
remove new_device_manager from lab

### LINKED ISSUE #295: The new device manager (labs feature) verifies current session and doesn't work
Tested in dev backend. 

Reproduce :
 - Activate feature in settings, in Labs/experimental (feature added in element-web 1.11.12)
 - Shows an extra section in settings, which prompts users to verify sessions : 
image

Without cross-signing, can users verify sessions ? Should we just hide this until we have cross-signing deployed ?

odelcroi:
Remove this new device manager from the lab, it could deployed and reworked when the cross signing is roll out

Capture d’écran 2023-01-16 à 16 21 54

### DIFF
--- a/config.dev.json
+++ b/config.dev.json
@@ -36,7 +36,8 @@
     "showLabsSettings": false,
     "features": {
         "feature_thread": true,
-        "feature_video_rooms" : false
+        "feature_video_rooms" : false,
+        "feature_new_device_manager" : false
     },
     "default_federate": true,
     "default_theme": "light",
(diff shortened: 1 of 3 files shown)


estellecomment:
Tested on preprod on review app : 
 - the feature is not displayed in Labs, you cannot enable it. 
 - if you had enabled it previously, the "Sessions" tab does not appear any more in the settings. 
 The feature is still enabled though : on another frontend you can still see it. But that seems ok.

estellecomment (approved):

PR #389, opened 2023-01-23
More patch checks : missing entries in patches.json or in patch/ directory

estellecomment:
This is to avoid having patches that are listed in patches.json but there is no corresponding patch file present.
And the reverse as well.


estellecomment:
All the CI is broken, because the new check breaks, which is what we want :)

odelcroi (approved):

PR #410, opened 2023-02-07
xsss : skip screens, rewording to "synchronize"

### LINKED ISSUE #406: Adapt secure storage configuration screens to the target design
Study how this integration can be done for a beta and after for a final version: 
- [x] only texts
- [x] define custom panels

https://www.figma.com/file/3qu9dfXgLomL6u04mpaTip/Tchap---v4?t=jzWqp5foPkuMzqCk-0

odelcroi:
> First iteration for internal beta

# Notification : 
Capture d’écran 2023-02-06 à 18 10 55

# Copy key
Capture d’écran 2023-02-06 à 18 11 04

# Activate
Capture d’écran 2023-02-06 à 18 11 10

## Checklist

-   [ ] Linter and other CI checks pass
-   [ ] Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/vector-im/element-web/blob/develop/CONTRIBUTING.md))

### DIFF
--- a/patches/patches.json
+++ b/patches/patches.json
@@ -170,5 +170,14 @@
       "src/components/views/context_menus/RoomContextMenu.tsx",
       "src/components/views/dialogs/SpacePreferencesDialog.tsx"
     ]
+  },
+  "rewording-key-backup-configuration": {
+    "comments" : "rewording on key backup configuration",
+    "github-issue" : "https://github.com/tchapgouv/tchap-web-v4/issues/406",
+    "package": "matrix-react-sdk",
+    "files": [
+      "src/async-components/views/dialogs/security/CreateSecretStorageDialog.tsx",
+      "src/toasts/SetupEncryptionToast.ts"
+    ]
   }
 }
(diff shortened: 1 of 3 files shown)


areox-net (changes requested):
@odelcroi I see you have already moved to the next step! 😃  

I think it works pretty well. The only thing that is a little odd at this point is"Activer ma sauvegarde" as the final button. In this scenario, I think it should read "Activer ma Clé".

areox-net:
@odelcroi It's great that we skip so many screens.
However, if when we click on "Synchroniser" from the small modal we are directly redirected to the "Clé de synchronisation", maybe the initial message in the small modal could be more specific and already mention "Clé de synchronisation".
"Synchroniser cet appareil" could only replace the case where we currently say "Vérifier cet appareil" if we need to differentiate the two cases. 

Screenshot 2023-02-07 at 15 08 25

odelcroi:
Interesting.
I'm not sure that people will be teased by an invite that says "Activez
votre clé de synchronisation". I mean, would they understand what it is
about?
However the underlying body message tells it, do you think it is enough?

areox-net:
We could have something between the two:

Screenshot 2023-02-07 at 16 06 22

aulamber (approved):

PR #431, opened 2023-02-21
428 change csss wording

aulamber:
See issue https://github.com/tchapgouv/tchap-web-v4/issues/428


areox-net (approved):

estellecomment:
I haven't followed the latest news in terms of wording. If Angelo is happy then it sounds good to me. 
Not reviewing !

areox-net:
@estellecomment basically, for the v1 of the external beta we decided to stick to the Element wording and flow, changing as little as possible. As usual, the idea is to push Element to improve their wording and flow instead of making hard-to-maintain patches in Tchap.

PR #438, opened 2023-02-27
Green confirmation flag for device verification

### LINKED ISSUE #432: Show the confirmation flag in green instead of blue
After a confirmation by QRCode the mobile app show a green shield and the web app show a blue shield.
There is a question that said : do you see the same shield ?
Should be no because the shield haven't the same color.

To Do : use green color for the shield (maybe it should be patched in Element)

estellecomment:
Fixes #432 

I used the same green as element's main accent color, as a green confirmation color.
It had been replaced by tchap's accent color, which is blue. 

Before : 
Screen Shot 2023-02-27 at 11 16 00 AM

After : 
Screen Shot 2023-02-27 at 11 25 30 AM


aulamber (approved):

aulamber:
Code et review app ok

PR #454, opened 2023-03-10
NE PAS MERGER : Experimentation activation des Espaces

jdauphant:


odelcroi:
@jdauphant can it be closed in favor of tchap lab?

jdauphant:
Yes, of course

Issue #445, opened 2023-03-01
Study how to deactivate setup xsss toast


odelcroi:
Study how to deploy a feature flag to setup a feature flag to deactivate this popup toast : 

<img width="495" alt="Capture d’écran 2023-03-01 à 17 45 32" src="https://user-images.githubusercontent.com/4077729/222206026-f9f26ee3-b129-4fb9-b9e4-24b31d1f5961.png">

Why : 
- deploy xsss silently
- let users configure it if they want to
- do not motivate them to do it

Maybe this line is a good start : 
https://github.com/matrix-org/matrix-react-sdk/blob/cf36acaa418d3a698cb7def4f5208faa1829a00e/src/DeviceListener.ts#L265

PR #455, opened 2023-03-13
Deactivate setup xsss toast, activate xsss ff in production

aulamber:
Cf. issue https://github.com/tchapgouv/tchap-web-v4/issues/445

Hide this setup toast:
Screenshot 2023-03-13 at 11 40 45

To reproduce:
- create a new account
- on first log in, create a private room. Then the setup toast which previously popped up with cross-signing on, doesn't. Like this:
Screenshot 2023-03-13 at 11 28 30

However, this other toast still pops-up on re-login (as it should; make sure that it does in your test):
Screenshot 2023-03-13 at 11 28 55

### DIFF
--- a/config.prod.json
+++ b/config.prod.json
@@ -108,7 +108,8 @@
         "feature_thread": true,
         "feature_video_rooms" : false,
         "feature_new_device_manager" : false,
-        "tchap_activate_cross_signing_and_secure_storage" : false
+        "tchap_activate_cross_signing_and_secure_storage" : false,
+        "tchap_disable_cross_signing_setup_toast": false
     },
     "default_federate": true,
     "default_theme": "light",
--- a/config.prod.json
+++ b/config.prod.json
@@ -108,8 +108,8 @@
         "feature_thread": true,
         "feature_video_rooms" : false,
         "feature_new_device_manager" : false,
-        "tchap_activate_cross_signing_and_secure_storage" : false,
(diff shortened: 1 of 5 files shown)


src/util/TchapUIFeature.ts
```diff
@@ -38,6 +38,19 @@ export default class TchapUIFeature {
         return isCrossSigningAndSecureStorageActive;
     };
 
+    /**
+     * This flag activates cross-signing and secure storage. But it also hides the xsss setup toast. 
+     * needs to be initialized to access the config.json (where the flag is set)
+     */
+        public static isCrossSigningAndSecureStorageActiveWithoutSetupToast = () : Boolean => {
```
odelcroi: Il me parait plus modulaire de faire une fonction pour récupérer uniquement la valeur du flag "tchap_disable_cross_signing_setup_toast". 
Au final c'est pas si important

config.prod.json
```diff
@@ -108,7 +108,8 @@
         "feature_thread": true,
         "feature_video_rooms" : false,
         "feature_new_device_manager" : false,
-        "tchap_activate_cross_signing_and_secure_storage" : false
+        "tchap_activate_cross_signing_and_secure_storage" : false,
```
odelcroi: pourrait-on activer tchap_activate_cross_signing_and_secure_storage à true en prod aussi, sinon ca complique les tests de beta interne

odelcroi (approved):

PR #456, opened 2023-03-14
Change wrong Personal Messages label when searching a person in tchap…

aulamber:
Cf. issue https://github.com/tchapgouv/tchap-web-v4/issues/448

See screenshots with modifications (from "Messages personnels" to "Personnes"):

Screenshot 2023-03-14 at 11 14 54
Screenshot 2023-03-14 at 11 14 49

### DIFF
--- a/src/i18n/strings/tchap_translations.json
+++ b/src/i18n/strings/tchap_translations.json
@@ -539,6 +539,11 @@
     "en": "Direct messages",
     "fr": "Messages directs"
   },
+  "Persons": {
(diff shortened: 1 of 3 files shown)


src/i18n/strings/tchap_translations.json
```diff
@@ -539,6 +539,11 @@
     "en": "Direct messages",
     "fr": "Messages directs"
   },
+  "Persons": {
+    "comments" : "Use persons to differenciate from Element label 'People' which we use in Tchap as 'Messages directs'",
```
odelcroi, line 543: Maybe we can rollback the modification we have made to "People". This way we avoid to introduce a new label

PR #479, opened 2023-03-20
Refactor patches.sh and add create script

jdauphant:
New script to merge, and new option to create patch.


scripts/PATCHES.md
```diff
@@ -0,0 +1,51 @@
...
+
+```
+./scripts/patche.sh create [patch_name] [package_name]
+```
+
+## Notes
+
+- Make sure to commit the changes to your patches folder after updating or creating patches.
+- The script assumes you're using yarn as your package manager. If you're using a different package manager, you'll need to modify the script accordingly.
+- Ensure you have Python installed on your system for the script to work correctly.
```
odelcroi, line 50: which version of python? no differnces?

odelcroi (changes requested):
I think it is great. Currently we base our workflow with the file patches.json and the related scripts makePatch.ts, apply_patches.sh, checkPatchFile.js, we need to study how to merge the previous and new scripts. 
Thos wofklow are described here : /patches/README.md

scripts/PATCHES.md
jdauphant: I have tried with python3

odelcroi:
closing for now. It would need more work to be compatible with how we work with yarn links

Issue #477, opened 2023-03-20
reset password : some buttons are not working


odelcroi:
4.1.7
Those buttons have not effect
<img width="466" alt="Capture d’écran 2023-03-20 à 15 44 56" src="https://user-images.githubusercontent.com/4077729/226376231-fefbe174-1061-40f9-b183-23ab67ae6185.png">

NicolasBuquet:
In fact, it is the DIV `<div role="button" tabindex="0" class="mx_AccessibleButton mx_AuthBody_resend-button mx_AccessibleButton_hasKind mx_AccessibleButton_kind_link">` which has its property `display: none`.

When `display` is not to `none`, the resend button is displayed : 

![image](https://user-images.githubusercontent.com/18608158/226411618-f3734905-bcf8-4dae-8678-ef0201f9815f.png)

odelcroi:
Not blocking but a annoying regression and quick to fix then P1

PR #499, opened 2023-03-27
Fix reset password unworking buttons

aulamber:
Cf. issue https://github.com/tchapgouv/tchap-web-v4/issues/477

When not working it looked like:
Capture d’écran 2023-03-27 à 16 56 11

And after fixing, it looks like this (and the buttons work):
Capture d’écran 2023-03-27 à 16 56 33

### DIFF
--- a/res/themes/tchap-common/css/_tchap_custom.pcss
+++ b/res/themes/tchap-common/css/_tchap_custom.pcss
@@ -1,9 +1,6 @@
 /* override components style */
 /* those overrides are imported in all themes of tchap */
-/* [TEMPORARY] hide "Sign in instead" button on password forgotten page because the redirection is currently broken */
-.mx_AuthBody.mx_AuthBody_forgot-password .mx_AccessibleButton {
-    display: none;
-}
+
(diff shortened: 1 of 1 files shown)


odelcroi:
> And after fixing, it looks like this (and the buttons work):

does the "Sign in instead" button mentionned from code comments work also? 
/* [TEMPORARY] hide "Sign in instead" button on password forgotten page because the redirection is currently broken */

aulamber:
> /* [TEMPORARY] hide

Oui le bouton fonctionne et il est bien affiché dans la page forgot_password ("Entrez votre adresse mail pour réinitialiser votre mot de passe").

odelcroi (approved):

odelcroi:
great,
a screenshot would have been nice and I would not have asked confirmation for that

aulamber:
> great, a screenshot would have been nice and I would not have asked confirmation for that

There was no screenshot for that modal because the issue was not related to the modal with the "Sign in instead" button, but to the confirmation modal ("Verifiez vos mails avant de continuer")

Issue #375, opened 2023-01-16
Check if patches directories match the contents of patches.json


estellecomment:
It seems to be possible to have dirs in patches which are not present in patches.json, and the scripts do not complain. 
This should not be possible, so make the scripts crash in this case.

odelcroi:
Could not reproduce, if a folder is added in patches/, the script complains if there is no matching entry in patches.json

PR #506, opened 2023-03-29
Make patches.json match the files actually modified in the patches

aulamber:
Cf. https://github.com/tchapgouv/tchap-web-v4/issues/375

[Warning]:
This PR does not fix the issue which was to create a script to automate the fact that files names in patches.json match the actual files in the patches.
This PR does it manually by removing the current useless files in patches.json.

But we will need in the future to automate this process.

### DIFF
--- a/patches/patches.json
+++ b/patches/patches.json
@@ -1,7 +1,10 @@
 {
     "are-external-users-allowed-in-room": {
         "package": "matrix-react-sdk",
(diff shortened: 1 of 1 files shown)


odelcroi:
I think the ticket was meant to change the script so that this situation does not happen again. It seems that you did it manually, this the first step to automate then.

odelcroi (approved):

Issue #436, opened 2023-02-23
XSSS : move xs "reinitialize" button to advanced section


estellecomment:
Example : "disable cross signing" button

<img width="341" alt="Capture d’écran 2023-01-24 à 14 53 10" src="https://user-images.githubusercontent.com/4077729/214312662-e79ff3ba-78df-4f3e-95a8-2aec490ff9bb.png">

Any others ?

estellecomment:
When cross-signing is activated, we see : 
<img width="634" alt="Screen Shot 2023-02-27 at 11 47 25 AM" src="https://user-images.githubusercontent.com/911434/221543850-90327a97-8224-4871-b623-3bc67bf639c6.png">

jdauphant:
I will keep it, I don't see good reason to remove the button

estellecomment:
@odelcroi You raised this problem originally. 
I also have doubts that we should remove this "Réinitialiser" button, since it could be helpful to fix some strange encryption situations.
What did you have in mind ? 

If we want to avoid having user click this button and mess up their session accidentally, we can move it inside the "Avancé" section which is hidden by default.

odelcroi:
> If we want to avoid having user click this button and mess up their session accidentally, we can move it inside the "Avancé" section which is hidden by default.

This is what I have in mind. We should not disturb UX with advanced buttons like this one. I don't see a use case where users should reinitialize the cross signing.

odelcroi:
The warning message is clear, users are not supposed to use this button, thus : 
- as the message is clear we can hope users will not finalize the process
- the button can be moved to the advanced section (retargeting this issue)
- the issue can be P3 as the risk of messing the state is low

<img width="750" alt="Capture d’écran 2023-03-01 à 11 09 49" src="https://user-images.githubusercontent.com/4077729/222108992-e5b1bc2e-d0b7-4aeb-ac87-6b93904a042e.png">

PR #509, opened 2023-03-30
UX issues: Change buttons I lost my recovery code I want a new one

aulamber:
Cf. https://github.com/tchapgouv/tchap-web-v4/issues/467

See also issue https://github.com/tchapgouv/tchap-web-v4/issues/436 and related PR: https://github.com/tchapgouv/tchap-web-v4/pull/510.

Before:
Capture d’écran 2023-03-30 à 11 07 30

After:
Capture d’écran 2023-03-30 à 11 07 51


Caroline-lawson:
👍  That should work

areox-net (changes requested):
All good except for the button wording, which is not consistent with this issue https://github.com/tchapgouv/tchap-product/issues/127

The button should be "Générer un nouveau code"

aulamber:
> All good except for the button wording, which is not consistent with this issue [tchapgouv/tchap-product#127](https://github.com/tchapgouv/tchap-product/issues/127)
> 
> The button should be "Générer un nouveau code"

Label modified. Is it ok with you @areox-net ? Can you validate the PR ? Thanks :)

areox-net (approved):

PR #512, opened 2023-03-30
Add error message when domain is not allowed on Tchap

### LINKED ISSUE #460: user register: improve error message when domain is not register
image (6)

- [ ] have a comprehensible message
- [ ] if it is an external user, then they may missing an invitation
- [ ] A link to Demarches Simplifiés should be provided to help users to follow the procedure to add the domain

odelcroi:
Before : 
see ticket #460 

After : 
Capture d’écran 2023-03-30 à 15 43 25


odelcroi:
- [x] What about external users that don't have an invit? test if they have the same message
- [ ] Yes, they do, maybe the message should talk to external users too
- [ ] maybe the link should go to https://tchap.beta.gouv.fr/

areox-net (changes requested):
The message is ok.
However, I would not use the blue color for the link. It could just be an underlined part of the red text.

areox-net (approved):

areox-net (approved):

areox-net (approved):

Issue #497, opened 2023-03-27
Rework securirty privacy settings


odelcroi:
### Your use case

Open security settings : 
![Capture d’écran 2023-03-27 à 16 01 11](https://user-images.githubusercontent.com/4077729/227962402-63b6f71e-0d94-405b-9e00-c2564e82ced1.png)

Would that be better to put first the Secure Backup section and then the device list

jdauphant:
@odelcroi what user problem do you want to solve ? can you explain in the ticket ?
Also, there is a new session manager that put "session" in other part of the settings

odelcroi:
Some feedbacks have been gathered to improve security configuration and understanding
- "Sauvegarde automatique des messages“  at the top in the settings
- Some rewording

PR #513, opened 2023-03-31
Change section order in security settings

aulamber:
Cf. issue https://github.com/tchapgouv/tchap-web-v4/issues/497

NB: Instead of creating a new patch, I realized that there was an existing patch with a purpose relatively similar to what I meant to do in this PR. The patch was `change-confusing-chiffrement-sub-session-in-parameters`. I made a combo out of the two, renaming the patch `change-sections-order-in-security-privacy-settings`.

Before:
Capture d’écran 2023-03-31 à 10 32 04

After:
Capture d’écran 2023-03-31 à 10 32 20


odelcroi:
what do you think? @areox-net

aulamber:
Todo: supprimer la section "Recherche de messages"

areox-net:
I think it's a good idea.
We should also rename the last section "Appareils connectés"

PR #524, opened 2023-04-04
Rework privacy settings (change order) in parameters

aulamber:
Cf. issue https://github.com/tchapgouv/tchap-web-v4/issues/497

NB: Instead of creating a new patch, I realized that there was an existing patch with a purpose relatively similar to what I meant to do in this PR. The patch was `change-confusing-chiffrement-sub-session-in-parameters`. I made a combo out of the two, renaming the patch `change-sections-order-in-security-privacy-settings`.

Before:
Capture d’écran 2023-03-31 à 10 32 04

After:
Capture d’écran 2023-04-04 à 16 26 07

### DIFF
--- a/patches/change-sections-order-in-security-privacy-settings/matrix-react-sdk+3.69.1.patch
+++ b/patches/change-sections-order-in-security-privacy-settings/matrix-react-sdk+3.69.1.patch
@@ -52,25 +52,38 @@ index 34b52e4..8085f20 100644
              </div>
          );
 diff --git a/node_modules/matrix-react-sdk/src/components/views/settings/tabs/user/SecurityUserSettingsTab.tsx b/node_modules/matrix-react-sdk/src/components/views/settings/tabs/user/SecurityUserSettingsTab.tsx
-index 9697af8..04ae847 100644
+index 9697af8..00fcc04 100644
 --- a/node_modules/matrix-react-sdk/src/components/views/settings/tabs/user/SecurityUserSettingsTab.tsx
 +++ b/node_modules/matrix-react-sdk/src/components/views/settings/tabs/user/SecurityUserSettingsTab.tsx
+@@ -287,7 +287,7 @@ export default class SecurityUserSettingsTab extends React.Component<IProps, ISt
+     public render(): React.ReactNode {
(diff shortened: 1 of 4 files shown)


odelcroi:
converted to draft as we are not quite sure about what the target should be
@areox-net the idea is to promote the secure backup, but the device view is also important, I would say less important though

areox-net:
I think I already discussed this with @aulamber during a call. 
@odelcroi I think it makes sense to change the order to make the XSS easier to find for the users.

odelcroi:
I've made a change so that : 
- users get to secure backup instantly
- devices list is upper than cross signing, because more important
- cross signing and the rest is left bottom with the other encryption stuff

@areox-net The elements on the web page are very spaced out because I was unable to reduce the margins in a not intrusive way. If you find it blocking, I'll go into it

Capture d’écran 2023-04-17 à 18 48 06

areox-net (approved):
I think this order works pretty well. 
The spacing between the elements is not an issue. On the contrary it makes it easier for the users to see the 3 blocs.

areox-net:
@odelcroi At this point, if XS is the only element left in "Chiffrement", couldn't we get rid of this title (which is not clear for users) and just have "Signature croisée" as the section title like you did with "Sauvegarde automatique..." at the top of the page ?

odelcroi:
> @odelcroi At this point, if XS is the only element left in "Chiffrement", couldn't we get rid of this title (which is not clear for users) and just have "Signature croisée" as the section title like you did with "Sauvegarde automatique..." at the top of the page ?

In fact there is two things in Chiffrement : tchap keys and signature croisée. I will merge as is. feel free to reopen an issue if needed

areox-net:
> > @odelcroi At this point, if XS is the only element left in "Chiffrement", couldn't we get rid of this title (which is not clear for users) and just have "Signature croisée" as the section title like you did with "Sauvegarde automatique..." at the top of the page ?
> 
> In fact there is two things in Chiffrement : tchap keys and signature croisée. I will merge as is. feel free to reopen an issue if needed

I see. I forgot about that. Let's keep like that then.

PR #552, opened 2023-04-18
New custom light theme

### LINKED ISSUE #544: Do not use red text when the user is mentioned.

areox-net:
Here is a PR to test the custom light theme.

### DIFF
--- a/res/themes/tchap-light/css/_tchap_custom.pcss
+++ b/res/themes/tchap-light/css/_tchap_custom.pcss
@@ -8,49 +8,98 @@
     color: red; //<- new attribute
 } */
 
-.mx_SpacePanel .mx_SpaceButton.mx_SpaceButton_home .mx_SpaceButton_icon:before {
-    background-color: #000091;
-}
+$color_1: #2E2F32;
+$color_2: white;
+$color_3: #68707d;
+$background-color_1: #000091;
+$background-color_2: transparent;
+$background-color_3: rgba(125,125,150,0.65);
+$background-color_4: #c7d5e4;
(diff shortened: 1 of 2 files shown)


odelcroi:
can run "yarn prettier -w res" to remove the lint error? This way, it will create the review app, and we can test it live

odelcroi:
@areox-net des screenshots seraient cool pour voir les changement

areox-net:
Here is a quick before/after. 

Before
Screenshot 2023-04-21 at 13 54 02

After
Screenshot 2023-04-21 at 13 53 55

odelcroi:
As we consider this theme as a "light", isn't the left bar too contrasted with the rest of the UI ? It looks too constrated to me

areox-net:
> As we consider this theme as a "light", isn't the left bar too contrasted with the rest of the UI ? It looks too constrated to me

I see your point, but the "light" theme is mostly defined by the background of the main containers. This sidebar is more like a vertical header (see Github header as a reference). In my opinion the contrast is justified.

odelcroi:
With a darker left bar : @areox-net ? 
Capture d’écran 2023-05-05 à 10 43 58

areox-net:
@odelcroi It's not bad, but I would not go so far on desaturation. The color start being too violet. I think #222777 could be a good compromise between the current accent color and what you suggest. With this color for the sidebar we could even keep the current accent color I think. 

Screenshot 2023-05-05 at 11 21 08

odelcroi (approved):

areox-net:
All good for me.

Issue #528, opened 2023-04-11
study rebase 1.11.29


odelcroi:
Updates:
- Show room create icon if "UIComponent.roomCreation" is enabled: https://github.com/matrix-org/matrix-react-sdk/pull/10364/files
- Show "Invite" menu option if "UIComponent.sendInvites" is enabled: https://github.com/matrix-org/matrix-react-sdk/pull/10363

- Notifications: inline error message on notifications saving error: https://github.com/matrix-org/matrix-react-sdk/pull/10288

- Improved a11y for Field feedback and Secure Phrase input: https://github.com/matrix-org/matrix-react-sdk/pull/10320

- Fix long display name overflowing reply tile on IRC layout: https://github.com/matrix-org/matrix-react-sdk/pull/10343

I did not find PR that would impact our code negatively.

When trying to do a rebase, I also noticed that something was added in .github/workflows/static_analysis.yaml, related to typescript:

```
    tsc-strict:
        name: Typescript Strict Error Checker
        if: github.event_name == 'pull_request'
        runs-on: ubuntu-latest
        permissions:
            pull-requests: read
            checks: write
        steps:
            - uses: actions/checkout@v3

            - name: Install Deps
              run: "scripts/tchap/install-yarn-linked-repositories.sh"

            - name: Get diff lines
              id: diff
              uses: Equip-Collaboration/diff-line-numbers@df70b4b83e05105c15f20dc6cc61f1463411b2a6 # v1.0.0
              with:
                  include: '["\\.tsx?$"]'

            - name: Detecting files changed
              id: files
              uses: futuratrepadeira/changed-files@96d5fd702a6479d573287ef07381ad59acc390ed # v4.0.0
              with:
                  repo-token: ${{ secrets.GITHUB_TOKEN }}
                  pattern: '^.*\.tsx?$'

            - uses: t3chguy/typescript-check-action@main
              with:
                  repo-token: ${{ secrets.GITHUB_TOKEN }}
                  use-check: false
                  check-fail-mode: added
                  output-behaviour: annotate
                  ts-extra-args: "--strict --noImplicitAny"
                  files-changed: ${{ steps.files.outputs.files_updated }}
                  files-added: ${{ steps.files.outputs.files_created }}
                  files-deleted: ${{ steps.files.outputs.files_deleted }}
                  line-numbers: ${{ steps.diff.outputs.lineNumbers }}
```

PR #566, opened 2023-04-25
Upgrade element web v1.11.29

aulamber:
Cf. issue https://github.com/tchapgouv/tchap-web-v4/issues/528

Upgrade + add documentation for merge patches in patches/README


package.json
```diff
@@ -67,7 +70,6 @@
         "test:cypress:open": "cypress open",
         "coverage": "yarn test --coverage",
         "analyse:unused-exports": "node ./scripts/analyse_unused_exports.js",
-        "postinstall": "./scripts/tchap/apply_patches.sh",
```
odelcroi: Ceci peut avoir un serieux impact sur notre systeme de patch

res/apple-app-site-association
```diff
@@ -3,7 +3,12 @@
         "apps": [],
         "details": [
             {
-                "appID": "7J4U792NQT.im.vector.app",
+                "appIDs":[
```
odelcroi: note : investigate what is it about

scripts/layered.sh
```diff
@@ -24,15 +24,15 @@ export PR_REPO=element-web
 node_modules/matrix-react-sdk/scripts/fetchdep.sh matrix-org matrix-js-sdk
 pushd matrix-js-sdk
 yarn link
-yarn install --pure-lockfile
+yarn install --frozen-lockfile
```
odelcroi, line 27: what is the difference?

src/i18n/strings/pl.json
```diff
@@ -0,0 +1,32 @@
+{
+    "Dismiss": "Zamknij",
```
odelcroi: we can delete this not needed file

src/i18n/strings/zh_Hant.json
```diff
@@ -0,0 +1,32 @@
+{
+    "Dismiss": "關閉",
```
odelcroi: we can delete this not needed file

src/vector/app.tsx
```diff
@@ -168,16 +168,14 @@ async function verifyServerConfig(): Promise<IConfigOptions> {
...
-                ),
+            throw new UserFriendlyError(
+                "Invalid configuration: can only specify one of default_server_config, default_server_name, " +
+                    "or default_hs_url.",
             );
         }
         if (incompatibleOptions.length < 1) {
             // noinspection ExceptionCaughtLocallyJS
-            throw newTranslatableError(_td("Invalid configuration: no default server specified."));
+            throw new UserFriendlyError("Invalid configuration: no default server specified.");
```
odelcroi, line 178: this new error looks interesting for us : will it help at translate server errors?

odelcroi (changes requested):

package.json
odelcroi: En effet, j'ai testé la review app, aucun patch n'est appliqué

aulamber: Done

scripts/layered.sh
aulamber: `yarn install --pure-lockfile` installs dependencies based only on the yarn.lock file. If the file is missing or outdated, Yarn will fail the installation. This command ensures that only dependencies listed in the lockfile will be installed, without performing any updates on them.

On the other hand, `yarn install --frozen-lockfile` also installs dependencies based on the lockfile, but it also prevents any modifications to the lockfile or the node_modules folder. If any changes are detected in the lockfile, Yarn will fail the installation.

The main difference is that `yarn install --frozen-lockfile` prevents any changes to the lockfile or node_modules folder, while `yarn install --pure-lockfile` only installs dependencies based on the lockfile without any updates.

res/apple-app-site-association
aulamber: The "apple-app-site-association" file is used to configure the association of our web app with specific iOS app IDs for universal links.
In earlier versions of iOS, it was only possible to associate one app ID with a domain for universal links. However, in newer versions of iOS, it's now possible to associate multiple app IDs with a domain.
As a result, the "apple-app-site-association" file now uses the "appIDs" key instead of the "appID" key to allow for the specification of multiple app IDs associated with a domain.

src/vector/app.tsx
aulamber: As a reminder the previous function used was:
```
export function newTranslatableError(message: string, variables?: IVariables): ITranslatableError {
    const error = new Error(message) as ITranslatableError;
    error.translatedMessage = _t(message, variables);
    return error;
}
```


Now it is:

```
export class UserFriendlyError extends Error {
    public readonly translatedMessage: string;

    public constructor(message: string, substitutionVariablesAndCause?: IVariables & ErrorOptions) {
        const errorOptions = {
            cause: substitutionVariablesAndCause?.cause,
        };
        // Prevent "Could not find /%\(cause\)s/g in x" logs to the console by removing
        // it from the list
        const substitutionVariables = { ...substitutionVariablesAndCause };
        delete substitutionVariables["cause"];

        // Create the error with the English version of the message that we want to show
        // up in the logs
        const englishTranslatedMessage = _t(message, { ...substitutionVariables, locale: "en" });
        super(englishTranslatedMessage, errorOptions);

        // Also provide a translated version of the error in the users locale to display
        this.translatedMessage = _t(message, substitutionVariables);
    }
}
```

The interesting aspects are that it will help with translating server errors, since it has a translatedMessage property that can be used to store the translated version of the error message to be displayed to the user. By using the _t function to generate the translated version of the error message, we can ensure that the error message is properly localized based on the user's language.

An example of its use:

```
throw new UserFriendlyError("Failed to load data. Status code: {status}", {
        status: apiError.status,
        cause: apiError.cause
      });
```
 
We can add more details to the error.

package.json
```diff
@@ -30,6 +30,9 @@
     "scripts_comments": {
         "build:jitsi": "tchap don't need that, deactivate?"
     },
+    "matrix_i18n_extra_translation_funcs": [
```
odelcroi, line 33: what is it? it seems interesting

patches/cross-signing-ui.log
```diff
@@ -0,0 +1,19 @@
+yarn run v1.22.19
+$ patch-package matrix-react-sdk
```
odelcroi: this file should be removed

patches/cross-signing-ui/matrix-react-sdk+3.70.0.patch
```diff
@@ -91,21 +91,15 @@ index 3e48739..07207de 100644
...
                      {_t(
                          "Store your Security Key somewhere safe, like a password manager or a safe, " +
-                             "as it's used to safeguard your encrypted data.",
-                     )}
-                 </p>
-+
-                 <div className="mx_CreateSecretStorageDialog_primaryContainer mx_CreateSecretStorageDialog_recoveryKeyPrimarycontainer">
-                     <div className="mx_CreateSecretStorageDialog_recoveryKeyContainer">
-                         <div className="mx_CreateSecretStorageDialog_recoveryKey">
+@@ -756,6 +792,7 @@ export default class CreateSecretStorageDialog extends React.PureComponent<IProp
```
odelcroi, line 94: not sure about this patch update?

odelcroi: Ok, get it, the blank line at line 97 was added by the patch. It wasn't necessary, it's ligher now

odelcroi (changes requested):

package.json
odelcroi: https://github.com/matrix-org/matrix-react-sdk/pull/7377

aulamber: I haven't found other occurrences of `matrix_i18n_extra_translation_funcs` in the matrix code, but it looks like it allows us to use the new error generating method `UserFriendlyError` which we discussed in a previous comment. And given the array in the package.json, we should now be able to use more than one error generating method :)

patches/cross-signing-ui.log
aulamber: Done

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -1,4 +1,3 @@
-blank_issues_enabled: false
```
odelcroi: can we keep that, our templates are not stable, so having a blank issue can be usefull

odelcroi (changes requested):

.github/ISSUE_TEMPLATE/config.yml
aulamber: Done

odelcroi (approved):

PR #568, opened 2023-04-25
Update user and avatar colors for Light theme

areox-net:
Here is a suggestion for issue https://github.com/tchapgouv/tchap-product/issues/152 

Actually, we can keep colors for users inside a room, while still using a blue-grey for the avatar background in the room list.


areox-net:
Here is a preview of the colors for user names in a room
Screenshot 2023-04-26 at 18 39 38

jdauphant (approved):

odelcroi (approved):

PR #584, opened 2023-05-04
Move js tchap files to a tchap folder

odelcroi:
Move all js files into : src/tchap/...

Extract tchap custom files to a dedicated folder : 
- [x] js files
- [ ] css files (outscoped from this PR)
- [x] test files
- [ ] workflow files  (outscoped from this PR)
- [x] script files


jdauphant:
Nice

PR #594, opened 2023-05-10
Fixes and improvements to custom CSS for Light Theme

areox-net:
This commit improves, fixes and extends the previous custom CSS code seen in #552 

Some CSS styles were !important in original CSS and were not previously taken into account.


areox-net:
Screenshot 2023-05-10 at 16 36 12

odelcroi (approved):

PR #664, opened 2023-08-29
Remove python from merge-patches script

estellecomment:
So that we don't add python as a dependency on dev laptops. 
Replaced the python line by an ugly complicated grep thing.


estellecomment:
hopefully made unnecessary if we move away from patches... 🤞

PR #665, opened 2023-08-30
Add feature flag based on homeserver

odelcroi:
Tested : 
- [x] feature shows when homeserver is in the list 
- [x] feature does not show when homeserver is not in the list 
- [x] feature does no show when no property is set

### DIFF
--- a/config.prod.json
+++ b/config.prod.json
@@ -186,7 +186,7 @@
     "permalink_prefix": "https://tchap.gouv.fr",
     "custom_translations_url_comments": "To add translations, edit /src/i18n/strings/tchap_translations.json. It gets copied over to webapp/i18n for serving.",
     "custom_translations_url": "/i18n/tchap_translations.json",
-    "tchap": {
+    "tchap_features": {
         "feature_email_notification": []
     }
 }
--- a/config.prod.json
+++ b/config.prod.json
@@ -186,7 +186,7 @@
     "permalink_prefix": "https://tchap.gouv.fr",
     "custom_translations_url_comments": "To add translations, edit /src/i18n/strings/tchap_translations.json. It gets copied over to webapp/i18n for serving.",
     "custom_translations_url": "/i18n/tchap_translations.json",
-    "tchap":{
+    "tchap": {
         "feature_email_notification": []
     }
 }
--- a/config.prod.json
+++ b/config.prod.json
@@ -185,5 +185,8 @@
     },
     "permalink_prefix": "https://tchap.gouv.fr",
     "custom_translations_url_comments": "To add translations, edit /src/i18n/strings/tchap_translations.json. It gets copied over to webapp/i18n for serving.",
-    "custom_translations_url": "/i18n/tchap_translations.json"
+    "custom_translations_url": "/i18n/tchap_translations.json",
+    "tchap":{
+        "feature_email_notification": []
+    }
 }
--- a/patches/add-feature-flag-based-on-homeserver/matrix-react-sdk+3.71.1.patch
+++ b/patches/add-feature-flag-based-on-homeserver/matrix-react-sdk+3.71.1.patch
@@ -17,7 +17,7 @@ index b2e44f2..28ff388 100644
 +//end :tchap: 
 \ No newline at end of file
 diff --git a/node_modules/matrix-react-sdk/src/components/views/settings/Notifications.tsx b/node_modules/matrix-react-sdk/src/components/views/settings/Notifications.tsx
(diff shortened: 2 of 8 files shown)


patches/add-feature-flag-based-on-homeserver/matrix-react-sdk+3.71.1.patch
```diff
@@ -0,0 +1,42 @@
...
++import TchapUIFeature from "../../../../../../src/tchap/util/TchapUIFeature";
+ 
+ // TODO: this "view" component still has far too much application logic in it,
+ // which should be factored out to other files.
+@@ -703,8 +704,9 @@ export default class Notifications extends React.PureComponent<IProps, IState> {
+                         />
+                     </>
+                 )}
+-
+-                {emailSwitches}
```
estellecomment: would be nice to comment out this code with TCHAP comments, to avoid seeing it reappear accidentally later

src/tchap/util/TchapUIFeature.ts
```diff
@@ -24,4 +27,26 @@ export default class TchapUIFeature {
...
+            ..
+        }
+    */
+    public static isFeatureActiveForHomeserver(feature:string):boolean {        
+        
+        const activatedHomeserver:[string] = SdkConfig.get("tchap")?.[feature] || [];
+        
+        //compare if user is on a listed homeserver
+        const userHomeServer = MatrixClientPeg.getHomeserverName();
+        return activatedHomeserver.includes(userHomeServer);
```
estellecomment:
```suggestion
        const homeserversWithFeature:[string] = SdkConfig.get("tchap")?.[feature] || [];
        const userHomeServer = MatrixClientPeg.getHomeserverName();
        return homeserversWithFeature.includes(userHomeServer);
```

src/tchap/util/TchapUIFeature.ts
```diff
@@ -24,4 +27,26 @@ export default class TchapUIFeature {
      */
     public static activateClearCacheAndReloadAtVersion4 = true;
 
+    /** 
+        get list of homeserver where the feature should be activated from config.json
```
estellecomment:
```suggestion
        get list of homeservers where the feature should be activated from config.json
```

patches/add-feature-flag-based-on-homeserver/matrix-react-sdk+3.71.1.patch
```diff
@@ -0,0 +1,42 @@
...
+--- a/node_modules/matrix-react-sdk/src/IConfigOptions.ts
++++ b/node_modules/matrix-react-sdk/src/IConfigOptions.ts
+@@ -192,3 +192,12 @@ export interface ISsoRedirectOptions {
+     immediate?: boolean;
+     on_welcome_page?: boolean;
+ }
++
++
++//:tchap: add tchap specific fields
++export interface IConfigOptions {
```
estellecomment: I'm confused about this. There is already a IConfigOptions declared in the file, is it overridden ? 

What is this used for ? Validating the fields ? Can you add a comment to explain ?

config.prod.json
```diff
@@ -185,5 +185,8 @@
     },
     "permalink_prefix": "https://tchap.gouv.fr",
     "custom_translations_url_comments": "To add translations, edit /src/i18n/strings/tchap_translations.json. It gets copied over to webapp/i18n for serving.",
-    "custom_translations_url": "/i18n/tchap_translations.json"
+    "custom_translations_url": "/i18n/tchap_translations.json",
+    "tchap": {
```
estellecomment: I would call it `tchap_features`.
Or `features_tchap`, and place it next to `features` in the file.
What do you think?

patches/patches.json
```diff
@@ -217,5 +217,13 @@
         "files": [
             "src/components/views/rooms/MemberList.tsx"
         ]
+    },
+    "add-feature-flag-based-on-homeserver": {
```
estellecomment: `tchap-features` ? `tchap-features-from-config`?
(our patch names are generally super long, so I'm trying to reduce them)

estellecomment (approved):
This is cool !
Not tested by hand, only code-reviewed.

config.prod.json
odelcroi: I've chosen a generic name because we might use it for other matters than feature flagging. A nice fourre-tout for tchap things.

patches/patches.json
odelcroi: yes

PR #675, opened 2023-08-31
try : keep security user settings tab as before upgrade

odelcroi:
Problem : 
- SecurityUserSettingsTab has changed a lot (components instead of raw html in render methods)
- Devices and DevicesEntry has been removed

Try : 
- restore as it was before #boomer

image


odelcroi:
closed in favor of #678

PR #682, opened 2023-09-06
add base url to make rageshake on preprod working

mcalinghee:
This PR is to make rageshake working on preprod.
The base url is missing in the config.json for preprod.

### DIFF
--- a/config.preprod.json
+++ b/config.preprod.json
@@ -25,7 +25,7 @@
     "disable_login_language_selector": false,
     "disable_3pid_login": false,
     "brand": "Tchap",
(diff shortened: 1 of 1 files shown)


estellecomment (approved):

estellecomment:
Is it relevant to fix config.dev.json as well ? Or is there no server so it's useless ?

mcalinghee:
> Is it relevant to fix config.dev.json as well ? Or is there no server so it's useless ?

It would be relevant but as we are not using it yet, I prefer it goes in another PR until dev is ready for this.

PR #709, opened 2023-09-21
CI : upload build from build job, and use it in cypress job

estellecomment:


estellecomment:
Merging without review because this can only work on the default branch (develop_tchap). Since cypress is broken anyway, I can't break anything important. :)

PR #710, opened 2023-09-21
CI : download build from uploaded artifact for cypress job 

estellecomment:
First part of work done in : https://github.com/tchapgouv/tchap-web-v4/pull/709 (merged because it would only work on default branch 🤦‍♀️ )

This is the second part which makes the whole thing work.


estellecomment:
Merging without review again, since I need all changes to be in the default branch (develop_tchap) otherwise the changes are ignored. 
(this is a bit annoying)

PR #722, opened 2023-09-25
[experiment] Reusable build job

estellecomment:
Alternative way of running the cypress tests : 
Write a build job in a separate file (reusable job), that is called by the cypress workflow.

I thought that the build and cypress jobs would share env files, but they don't, so we need to use caches to get the results of the build for cypress. And also the files from yarn install. 

Overall it's not clear that the added complication is really solving anything.

(the changes to review are in the .github/workflow dir, the others are from other PRs, ignore them)


estellecomment:
that was fun, ok bye

PR #730, opened 2023-09-28
[do not merge] Cypress : import setup files from react-sdk

estellecomment:
Instead of copying them. That way we stay up to date for e2e tests.

TODO : replace copies by imports in :
 - [x] cypress folder
 - [x] cypress/support folder
 - [ ] cypress/plugins folder
 - [ ] cypress/fixtures folder
 - [ ] devDependencies : remove useless ones, and track which are tchap-specific vs element-web


estellecomment:
Many files have been deleted, so this PR has a lot of conflicts. Could be worth starting over from scratch.

PR #732, opened 2023-09-28
Only run Cypress on non-draft PRs

estellecomment:
The previous version of the file ran the cypress job when the PR changed state from draft to "ready to review". So it didn't run when the PR was created directly "ready for review".

This version should run when the state is "ready for review", regardless of how the PR got to that state !

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,7 +1,8 @@
 #Build the application
 name: Build without E2E tests
 on:
-    pull_request: {} # only on draft PRs, see "if:" below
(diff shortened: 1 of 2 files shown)


estellecomment:
Fixed both jobs to use the two mechanisms : the trigger (ready_for_review and converted_to_draft) for when the PR status changes, and the `if` condition to filter whether or not the job is run. 

Maybe this will work. Maybe not. That's life in the world of github actions. 

Merging without review, in a mixed emotional state of hopeful fear.

PR #760, opened 2023-10-06
E2e test for content scanner (2nd try)

estellecomment:
There are problems with content scanner in dev env. These tests do work on other envs.


estellecomment:
Blocked by #780

estellecomment:
This works locally with a backend in which the content scanner works. Good enough for now.

PR #774, opened 2023-10-11
Migrate translations to module

estellecomment:
Add a module containing translations. 
 - tchap_translations.json is still the same, it just moved.
 - the module is a yarn/npm module, but placed in out own files, not in another repo or package server. If we put it in a separate repo, we would need to make sure the version of the translations module matched the code in the patches+customisations, so that for any change you would need two separate PRs. We used to do this with matrix-react-sdk and it was awful.
 - Changes in module code are not picked up by webpack-dev-server, not sure how to do that. So you need to stop and restart the server to see any changes, which is quite long. 🤷‍♀️ 
 
Edit : modified the installer to no run yarn add, because this broke scalingo. See https://github.com/tchapgouv/tchap-web-v4/pull/776
 
Cypress still broken in CI, but runs locally (only on prod env though. Not sure why.)


estellecomment:
Ah zut la review app est pétée. Bon.

estellecomment:
The problem is coming from the yarn add that is run to install modules. I suspect scalingo wants yarn add to be run during install phase, and not after. 

image

estellecomment:
Merging without review because it's been there long enough !

PR #790, opened 2023-10-19
Scripts to write the deployment PR

estellecomment:
Should this be in tchap-infra? 
Maybe! 

Should this be in bash?
Maybe but I hate bash!

It could also make sense to rethink what to write in the tchap-infra files, I'm not sure the naming is great.
Also this counts on the release having been built correctly by script. If the different assets have different dates, it breaks.


odelcroi:
look like a good start. A little example of a command being run would be helpfull.

estellecomment:
There is one in the file comments : 
* Usage : node scripts/tchap/writeReleaseConfigFiles.js --current=4.3.9 --new=4.3.9 --beta=4.3.10 --tchapInfraDir="../tchap-infra"
Where you looking for it somewhere else ?

estellecomment:
approved ? not approved ? I'm already using it :)

PR #309, opened 2022-11-16
Add check-translations.sh - Tool to detect missing and useless strings

estellecomment:
See README.md added in PR for explanations.

PR #824, opened 2023-11-13
WIP : translation keys I think are not used in app

estellecomment:
I could be wrong for the longer ones, since I'm just doing text searches in code.
We should figure out why they are unused, it could be lost code. 

Conversely, it could be dangerous to delete strings that we are using, by mistake... 

Ideally using a script to do this would be best...
(I had started in https://github.com/tchapgouv/tchap-web-v4/pull/309)


estellecomment:
useless now that we have translation-checking scripts !

PR #856, opened 2023-12-13
Activate audio and video call based on featureflag

### LINKED ISSUE #844: activer les appels AUDIO sur web pour les users DINUM
- [x] Uniquement pour les DM
- [x] Uniquement les appels Audio 
- [ ] Tests liés aux appels

## RoomHeader has been reworked in Element 

`RoomHeader.tsx` and `LegacyRoomHeader.tsx`
the new RoomHeader can be activated with the option : 
```
    "features": {
        "feature_new_room_decoration_ui" : true
..

```

## Activating only the audio button only for DM for Dinum
1/
A patch can be made to filter the render of the button is room is a DM and the user is part of the homeserver dinum

What is the UX difference between both  : 

###  LegacyRoomHeader
image

### RoomHeader
- DM
image

- Room
image

## Throughts

- The new version is lighter
- It is way easy to customize RoomHeader.tsx than LegacyRoomHeader
- LegacyRoomHeader is flagged as deprecated in the code

> it seems to be a good idea to activate the new RoomHeader.tsx

## Questions
- [ ] what is the status of the new version in terms of development?
- [ ] is there dramatic UX impacts if we activate it? 
- [x] activate new room header in tchap labs

odelcroi:
See config files : 
https://github.com/tchapgouv/tchap-web-v4/blob/e1b718e098fde7ea48a76384c7ab7984686f1a44/config.dev.json

### With External users before : 
Capture d’écran 2023-12-14 à 14 48 12

### With External users after : 
Capture d’écran 2023-12-14 à 14 47 52

### With Audio call before : 
Capture d’écran 2023-12-14 à 14 48 28

### With Audio call after : 
Capture d’écran 2023-12-14 à 14 48 43


odelcroi:
@areox-net the caption "ouvert aux externes" was easier to set before the name of the room, do you see a problem?

areox-net:
@odelcroi It's not the best solution, but I understand it's better for rooms with long titles. 
Ideally, I'd put it **under** the room name in a Title+Subtitle logic.

NicolasBuquet:
@odelcroi : Shouldn't we leave video call out of scope in this PR?

odelcroi:
> @odelcroi It's not the best solution, but I understand it's better for rooms with long titles. Ideally, I'd put it **under** the room name in a Title+Subtitle logic.

In this new version, under the room title, there is the room descritpion

odelcroi:
> @odelcroi : Shouldn't we leave video call out of scope in this PR?

The video button is hidden with a feature flag, then we can display it progressively with no code modification

areox-net (approved):

PR #876, opened 2024-01-22
Fix cypress tests after new RoomHeader

### LINKED ISSUE #875: create-room cypress tests broken because of LegacyRoomHeader -> RoomHeader

estellecomment:
Fixes https://github.com/tchapgouv/tchap-web-v4/issues/875

We moved from LegacyRoomHeader to RoomHeader, which broke some tests, and found a regression! Fix it.

The regression was that "Ouvert aux externes" had disappeared in the RoomHeader.

Cypress tests run locally on prod backend.


estellecomment:
Merging without review : tests pass, that's the whole point.

PR #886, opened 2024-02-05
Move Tchap workflows (github actions) in separate dir

### LINKED ISSUE #761: Separate tchap github actions from element's
to staop them being merged with element's when we run upgrades.

estellecomment:
Fix https://github.com/tchapgouv/tchap-web-v4/issues/761

Will avoid conflicts with element workflows during upgrades.


estellecomment:
Merging in develop_tchap because I can't see if it works while it's in the branch.

PR #887, opened 2024-02-05
Upgrade/element web v1.11.53

estellecomment:
- [x] remove patches/Better-text-for-tab-switch if it's not needed any more
- [x] Change-sections-order-in-security-privacy-settings : some strings fixed ? 
- [x] export-room-members : button probably moved out of the header. Is that ok ? 
- [x] remove-fdroid : check that the fix works now the patch is removed
- [x] No-voip-mentions-in-onboarding : remove a custom string ? 
- [x] room-summary-card: we're not using the same markup as element. Should we ? Check that RoomSummaryCard is not broken.
- [x] this : 
image

- [x] this : 
image

- [x] this : 
image
https://github.com/tchapgouv/tchap-web-v4/issues/890

- [ ] Should we offer "Vérifier avec un autre appareil" on login ?

Also :
- [x] check translations
- [x] cypress : passed locally with hand-extended timeouts in various places. It's becoming really hard to run. Issue : https://github.com/tchapgouv/tchap-web-v4/issues/889
- [x] fix CI : disable element actions, enable tchap actions
- [x] CI tests

Note : new warning appears : "start value has mixed support, consider using flex-start instead"
This is a pcss warning. Ex: `align-self: start;` -> `align-self: flex-start;`
Ignoring because the worst that can happen is a bit messed up layout.

- [ ] "Labs"/Experimental est affiché dans les settings : 
image


.github/labels.yml
odelcroi, line 1: we don't want those new labels in our repo

odelcroi (approved):
lgtm

.github/labels.yml
odelcroi: removed

odelcroi:
I found a blocking bug : 
Go to parameters>about : panel closes with console error. It is blocking because users can not access to their tchap about version
Capture d’écran 2024-02-09 à 15 26 15

odelcroi (changes requested):
blocking bug found, see comments

estellecomment:
Bug :     "Labs"/Experimental est affiché dans les settings :
image

(yavait pas un e2e test qui le checkait ??)

odelcroi:
> Bug : "Labs"/Experimental est affiché dans les settings : image
> 
> (yavait pas un e2e test qui le checkait ??)

Je ne le vois pas dans la review app, où l'as tu vu? Sinon on merge et on fixe après, ca semble pas bloquant

PR #933, opened 2024-03-11
fix(room-icons #908): remove deprecated function and use event timeline with join rules attributes

### LINKED ISSUE #908: Some rooms are not in the 3 types (public, private, private with externs)
Several of tchap teams' rooms don't have an icon of one of the 3 types of rooms (public, private, private with externs)

image

Is it because they are in the Tchap space ? Space-related bug ? Or because they were created by another client ?

MarcWadai:
Fix #908

## Checklist

-   [ ] Linter and other CI checks pass
-   [ ] Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/element-hq/element-web/blob/develop/CONTRIBUTING.md))

## Deprecated method from matrix-js-sdk
- room.currentState is deprecated
- isRoomEncrypted is also deprecated because of matrix rust sdk

## What's the problem
- The problem that was causing the icon to no show up sometimes is because the event received from the room where sometimes undefined. I noticed this behavior, only on rooms that have where configured on invite mode
- My guess is that maybe some configuration changed on the backend, so the value returned by state key `im.vector.room.access_rules` are not the same anymore

## What changed
- One solution that was implemented here is to use join_rules instead of `im.vector.room.access_rules` 

## TODO
Still in draft because i think the best solution is to try fix the backend and check why we dont receive the same values anymore 

# UPDATE 
We don't mix join_rule and access_rules in tchap. The identified pb is that the `access_rule` event is not set when the rooms have been upgraded to version 9

### DIFF
--- a/src/tchap/@types/tchap.ts
+++ b/src/tchap/@types/tchap.ts
@@ -9,6 +9,8 @@ export enum TchapRoomType {
 export enum TchapRoomAccessRule {
     Unrestricted = "unrestricted", // accessible to externals
     Restricted = "restricted", // not accessible to externals
+    Public = "public",
+    Invite = "invite",
 }
 
 export interface TchapIAccessRuleEventContent {
--- a/src/tchap/util/TchapRoomUtils.ts
+++ b/src/tchap/util/TchapRoomUtils.ts
@@ -2,16 +2,15 @@
  * Tchap Room utils.
  */
 
-import { Room } from "matrix-js-sdk/src/matrix";
-import { MatrixClientPeg } from "matrix-react-sdk/src/MatrixClientPeg";
+import { EventTimeline, EventType, Room } from "matrix-js-sdk/src/matrix";
 
-import { TchapRoomAccessRule, TchapRoomAccessRulesEventId, TchapRoomType } from "../@types/tchap";
+import { TchapRoomAccessRule, TchapRoomType } from "../@types/tchap";
 
 export default class TchapRoomUtils {
     //inspired by https://github.com/tchapgouv/tchap-android/blob/develop/vector/src/main/java/fr/gouv/tchap/core/utils/RoomUtils.kt#L31
     //direct type is not handled yet
     static getTchapRoomType(room: Room): TchapRoomType {
-        const isEncrypted: boolean = this.isRoomEncrypted(room.roomId);
+        const isEncrypted: boolean = this.isRoomEncrypted(room);
         const tchapRoomAccessRule: TchapRoomAccessRule = this.getTchapRoomAccessRule(room);
         return this.getTchapRoomTypeInternal(isEncrypted, tchapRoomAccessRule);
     }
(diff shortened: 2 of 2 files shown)


src/tchap/util/TchapRoomUtils.ts
```diff
@@ -38,15 +37,19 @@ export default class TchapRoomUtils {
      * @returns string that matches of one TchapRoomAccessRule //todo or null? or empty?
      */
     static getTchapRoomAccessRule(room: Room): TchapRoomAccessRule {
-        return room.currentState.getStateEvents(TchapRoomAccessRulesEventId, "")?.getContent().rule;
+        // the state event TchapRoomAccessRulesEventId is working only when the rule access is invite -> maybe a change in the backend configuration ? needs to check
```
odelcroi, line 40: This is expected. RuleAccess == invite defines a private room. 
In Tchap the TchapRoomAccessRule defines a subcategory of a private room. If TchapRoomAccessRule is unrestricted, then the private room will be accessible to users belonging to the "externes" homeserveur (we usually referre to those users as 'les externes' aka les gens qui ne sont pas agents publics)

src/tchap/@types/tchap.ts
```diff
@@ -9,6 +9,8 @@ export enum TchapRoomType {
 export enum TchapRoomAccessRule {
     Unrestricted = "unrestricted", // accessible to externals
     Restricted = "restricted", // not accessible to externals
+    Public = "public",
```
estellecomment, line 12: whoa, probablement pas. Ca correspond aux valeurs de im.vector.room.access_rules donc à priori on va pas en rajouter.
(mais il manque `direct`, est-ce qu'on l'utilise plus ? C'est possible)

Public et Invite ca ressemble plutot aux valeurs de [m.room.join_rules](https://spec.matrix.org/latest/client-server-api/#mroomjoin_rules)

Lire cette doc, ca va donner du contexte :
https://www.notion.so/mercurial-timer-ec4/Sp-cificit-s-Tchap-vs-Element-d-tails-329cf4d20ce840a3b53e29dd2d76f0f0?pvs=4#7f020e610f1e45e29a8a2ad032d2ee1d

src/tchap/util/TchapRoomUtils.ts
estellecomment, line 40: hmmm. If some rooms have no value for TchapRoomAccessRulesEventId ("im.vector.room.access_rules") it means that they were not created by a tchap client. (or the tchap client was in an experimental phase, most likely web v4 in its early stages :) )

In this case, the room could have any type of configuration (if it was created by an element client for example), and I don't know if we can decide on one of the 3 tchap-types. 🤔

src/tchap/util/TchapRoomUtils.ts
```diff
@@ -38,15 +37,19 @@ export default class TchapRoomUtils {
...
+        // That's is why sometimes the last state events is undefined, here we are not directly using getJoinRule because of types compatibility
+        const lastEvent =  room.getLiveTimeline().getState(EventTimeline.FORWARDS)?.getStateEvents(EventType.RoomJoinRules)[0];
```
estellecomment, line 42: RoomJoinRules is "m.room.join_rules" (standard matrix)
TchapRoomAccessRules is "im.vector.room.access_rules" (tchap specific)
Two different things ! See doc : https://www.notion.so/mercurial-timer-ec4/Sp-cificit-s-Tchap-vs-Element-d-tails-329cf4d20ce840a3b53e29dd2d76f0f0#84582964ab474317a2dc0e36f27c66bf

estellecomment:
Thanks for updating the deprecated functions !

PR #948, opened 2024-03-25
Fix build validated config from discovery tchap utils

MarcWadai:
## Description

Because of an update on the OIDC discovery implementation, a function that we used in Tchap went from sync to async. Cf this [PR](https://github.com/element-hq/element-web/pull/27047) in element-web.

That's why in this PR, I update the method and the different patches that use it.

## Checklist

-   [x] Linter and other CI checks pass
-   [ ] Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/element-hq/element-web/blob/develop/CONTRIBUTING.md))

### DIFF
--- a/src/tchap/util/TchapUtils.ts
+++ b/src/tchap/util/TchapUtils.ts
@@ -5,6 +5,7 @@ import { ValidatedServerConfig } from "matrix-react-sdk/src/utils/ValidatedServe
 import { findMapStyleUrl } from "matrix-react-sdk/src/utils/location";
 
 import TchapApi from "./TchapApi";
+import { ClientConfig } from "~tchap-web/yarn-linked-dependencies/matrix-js-sdk/src/autodiscovery";
 
 /**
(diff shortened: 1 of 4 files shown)


src/tchap/util/TchapUtils.ts
```diff
@@ -107,8 +108,8 @@ export default class TchapUtils {
      * @param
      * @returns
      */
-    static makeValidatedServerConfig = (serverConfig): ValidatedServerConfig => {
-        const discoveryResult = {
+    static makeValidatedServerConfig = async (serverConfig: Record<string, any>): Promise<ValidatedServerConfig> => {
```
estellecomment, line 111: rhooo le petit typage gratuit c'est beau

estellecomment (approved):
relu le code sans testé. 
Looks good, good catch !

estellecomment:
(fix linter probably in the main PR ? and force-merge this one without linter)

PR #962, opened 2024-04-04
Remove green from theme

### LINKED ISSUE #961: Theme regression ? back to element green
image

Not sure since which commit, did not investigate, probably since element upgrade ?

estellecomment:
Fix https://github.com/tchapgouv/tchap-web-v4/issues/961

Element's green colors reappeared after upgrade to 1.11.59 - 1.11.60 - 1.11.61, because of the addition of Compound to manage colors. 

This PR changes the minimum required to remove the green to respect the Tchap blue visual identity, and doesn't think further about whether we should change the theming more.

Testing : click everywhere to look for any green which should be turned back to blue :)

### DIFF
--- a/res/themes/tchap-light/css/_tchap_custom_vars.pcss
+++ b/res/themes/tchap-light/css/_tchap_custom_vars.pcss
@@ -31,6 +31,18 @@ body {
     --username-colors_6: #3f81e3;
     --username-colors_7: #000091;
 
+    /* Compound colors. Compound is a dependency for matrix-react-sdk, which manages colors (and maybe other style things ?) */
+    --cpd-color-bg-accent-rest: var(--accent);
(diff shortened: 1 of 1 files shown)


MarcWadai:
Good catch ! I think the dark theme is a little broken too. Does tchap have specification for this theme ? 
The switch buttons are still green and the sidebar menus became light gray 
image

MarcWadai (approved):

estellecomment:
image

Found one more !! Will add to PR before merge

estellecomment:
@MarcWadai yes the dark theme was not done in the first version, to gain time. And then it was never done. :) 
Filed https://github.com/tchapgouv/tchap-web-v4/issues/967

estellecomment:
> Found one more !! Will add to PR before merge

ah non false alarm :)

Issue #795, opened 2023-10-31
Add "Enable secure backup" in the steps in the welcome screen


estellecomment:
The welcome screen is displayed on firs tconnection, and when no rooms are selected :

<img width="1050" alt="image" src="https://github.com/tchapgouv/tchap-web-v4/assets/911434/740dd351-1259-4db7-ba26-a1dd38843bab">

Also accessible by clicking "Welcome" in the sidebar : 
<img width="444" alt="image" src="https://github.com/tchapgouv/tchap-web-v4/assets/911434/a44db874-3824-4fd4-baa9-ba647e68ee80">

This would allow us to nudge users a bit...

estellecomment:
@areox-net et @Caroline-lawson ca peut vous interesser

areox-net:
@estellecomment 100% d'accord.

estellecomment:
See DeviceListener and SetupEncryptionToast for implementation ideas.

estellecomment:
Also relevant : force activation of sauvegarde automatique : https://github.com/tchapgouv/tchap-web-v4/issues/807
If it works it could make this issue useless... Investigate.

estellecomment:
Issue filed in element-web as well : https://github.com/vector-im/element-meta/issues/2184

estellecomment:
Proposed design : 

<img width="1029" alt="image" src="https://github.com/tchapgouv/tchap-web-v4/assets/911434/d7b1a52e-ffd2-45f6-8b7b-c3afd75e21f3">

@areox-net any thoughts, especially about : 
 - text used ? 
 - should we remove steps ? 6 feels like too much, information overload. "Paramétrer votre profil" is the least useful for us I think, since there is only the picture to set.
 - position of this new step in the list ?

areox-net:
@estellecomment I will have these 4 items, in this order:

- Sauvegardez vos messages automatiquement
- Ajouter une photo de proil
- Trouvez et inviter vos collégues
- Téléchargez l'application mobile

Secondary texts are ok.

We don't need "compte créé and notifications"

estellecomment:
Thanks for the input !

Caroline-lawson:
For the text, verbs should either all be ER or EZ - I recommend ER so it is less imperative:

Sauvegarder vos messages automatiquement
Ajouter une photo de profil
Trouver et inviter vos collègues
Télécharger l'application mobile

PR #978, opened 2024-04-18
Reorder onboading and add secure storage task

MarcWadai:
## Issue
https://github.com/tchapgouv/tchap-web-v4/issues/795

## before
image

## after
image

## How to review 
- [ ] Check the onboarding display, all buttons should work
- [ ] Enable secure storage should be first
- [ ] His check icon should be green if secure storage is set
- [ ] His check icon should be gray if no secure storage is set

## TODO
- [x] add button and task
- [x] review text with design and product
- [x] open the dialog when button is clicked
- [x] set task as complete when backup is enabled
- [x] unit test


areox-net (changes requested):
Je pense qu'on peut changer la dernière action en 

"Télécharger l'application mobile"

areox-net (approved):

PR #982, opened 2024-04-23
fix(rageshake): reallow customapp field

MarcWadai:
issue https://github.com/tchapgouv/tchap-web-v4/issues/920 (small addition)

This will unbreak the crisp triage bot, which was using the customApp field (indirectly, taking the value from uisi_autorageshake_app in config.json).

## Check
- [x] modify tests


NicolasBuquet (approved):
LGTM

PR #995, opened 2024-05-02
feat(voip): hide video button when feature not active on homeserver

### LINKED ISSUE #952: Voip : video calls are possible from web to web
Tested on 4.4.2, between two DINUM accounts (voip is deployed for DINUM only, and should be audio-only, no video)

When making a call : 

 - from web to web : Only audio call button is shown, but once in the call, the "turn on video"  button is present. I can turn the call into a video call. I can also share my screen.
 
 
Screen Shot 2024-03-26 at 2 41 14 PM

Screen Shot 2024-03-26 at 2 41 36 PM

 -  from android to web : Only audio call button is shown. Once in the call, the video button is not present. I cannot turn the call into a video call.

 - from web to android : Only audio call button is shown. Once in the call, video button not present.

MarcWadai:
fix https://github.com/tchapgouv/tchap-web-v4/issues/952

## Description 
When in a call, the video button was visible from web to web call even if the feature was not activated in the config.
Added a check in the component to hide the button depending on the config

## Test
- [x] unit tests for showing and hiding the button depending on the config
- [x] manual test when calling and activating/deactivating the homeserver feature


NicolasBuquet (approved):
LGTM

PR #1051, opened 2024-06-25
Bump to 4.5.9

MarcWadai:
## Changes 
- Update version to 4.5.9, which is only build fix for version 4.5.8. So we don't update yarn lock and minor dependencies


NicolasBuquet (approved):
LGTM!

PR #1234, opened 2025-02-05
fix: remove tooltip on TAC

### LINKED ISSUE #1233: Le bouton centre d'activité des fils de discussion ne marche plus
Depuis la version 4.13.0, ce bouton n'est plus cliquable (rien ne se passe lorsque l'on clique dessus)

Malgré qu'aucun changement n'ai été apporté a cette section et que la même version de compound soit utilisé.

MarcWadai:
fixes https://github.com/tchapgouv/tchap-web-v4/issues/1233

Removing the tooltip fixes the issue... But the cause of it was still not found, so this is a hack

### DIFF
--- a/patches/tchap-modifications.json
+++ b/patches/tchap-modifications.json
@@ -193,5 +193,11 @@
         "files": [
             "src/hooks/room/useRoomCall.tsx"
         ]
+    },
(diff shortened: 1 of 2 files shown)


NicolasBuquet (approved):
It's good for me.

PR #1372, opened 2025-09-01
Fix Cross-window communication with unrestricted target origin

odaysec:
fix this issue, we should replace the `'*'` target origin in the `postMessage` call with a strict, explicit origin that matches the iframe's actual origin.  
This involves:
- Locating the origin corresponding to the iframe's `src` attribute (`"usercontent/"`). If it is always same-origin (relative URL), we can use `window.location.origin`.  
- In the send call, replace `'*'` with this origin.
- The change should be made within `src/utils/FileDownloader.ts`, directly where the `postMessage` call occurs.  
- No new imports are required: `window.location.origin` is a standard property.  
- If in any scenario the iframe's origin differs (e.g., constructed from a configurable URL), additional logic may be needed to accurately determine the origin.  
But for the current code shown (`src = "usercontent/"`), it is same-origin.

## Checklist

- [x] Tests written for new code (and old code if feasible).
- [x] New or updated `public`/`exported` symbols have accurate [TSDoc](https://tsdoc.org/) documentation.
- [x] Linter and other CI checks pass.
- [x] I have licensed the changes to Element by completing the [Contributor License Agreement (CLA)](https://cla-assistant.io/element-hq/element-web)

### DIFF
--- a/src/utils/FileDownloader.ts
+++ b/src/utils/FileDownloader.ts
@@ -90,7 +90,7 @@ export class FileDownloader {
                 download: name,
                 auto: autoDownload,
             },
-            "*",
+            window.location.origin,
         );
     }
 }


MarcWadai:
Thanks a lot for taking the time to open a PR 🙏. Tchap web is a soft fork of Element web, and this code doesn't touch specifics fonctionnalities about Tchap. So we recommend you to open it directly on https://github.com/element-hq/element-web so that all the community can benefits from it, including Tchap when synching with Element web.

PR #1373, opened 2025-09-01
Fix Client-side cross-site scripting

odaysec:
The correct and recommended way to prevent this vulnerability is to validate the value of `redirect_uri` before placing it in a sensitive attribute like `action`. The safest approach is to allow only a predefined list of redirect URIs (a whitelist), or, at minimum, to ensure that it is a relative path or matches the site's own origin, thereby preventing redirection to arbitrary external domains. For a demonstration server, you can restrict `redirect_uri` to relative URLs beginning with `/`, or check that the protocol and host match the current page's origin. 

To implement this, edit the relevant `<script>` in `auth.html` (lines 26-32) such that before setting the `action` attribute, the `redirect_uri` value is:
- Checked for presence (non-null)
- Validated to be a same-origin URL (using the URL API), or restricted to starting with `/`
- If invalid, either do not set the action, or set it to a safe default and optionally log or display an error.

## Checklist

- [x] Tests written for new code (and old code if feasible).
- [x] New or updated `public`/`exported` symbols have accurate [TSDoc](https://tsdoc.org/) documentation.
- [x] Linter and other CI checks pass.
- [x] I have licensed the changes to Element by completing the [Contributor License Agreement (CLA)](https://cla-assistant.io/element-hq/element-web)

### DIFF
--- a/playwright/plugins/oauth_server/res/oauth/auth.html
+++ b/playwright/plugins/oauth_server/res/oauth/auth.html
@@ -27,7 +27,27 @@ Mostly, it just redirects back to the `redirect_uri` in the query params.
             // process the query params, and set up the form
             const urlParams = new URLSearchParams(window.location.search);
             console.log("Test OAuth page: query params:", new Map(urlParams.entries()));
-            document.getElementById("auth_form").action = urlParams.get("redirect_uri");
+            // Set form action only if redirect_uri is a same-origin or relative path
+            const redirectUri = urlParams.get("redirect_uri");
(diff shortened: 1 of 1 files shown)


MarcWadai:
Thanks a lot for taking the time to open a PR 🙏. Tchap web is a soft fork of Element web, and this code doesn't touch specifics fonctionnalities about Tchap. So we recommend you to open it directly on https://github.com/element-hq/element-web so that all the community can benefits from it, including Tchap when synching with Element web.

PR #1403, opened 2025-09-23
define a port for local webserver

odelcroi:
For local testing, it is easier to define a web port mapping

(could be defined in .env?)


MarcWadai (approved):
Should be enough :) :+1:

PR #1472, opened 2025-12-08
TCHAP(desktop): remove stronghold and init keyring

MarcWadai:
## Changes
- Remove stronghold 
- Initialize keyring (use system keychain to store keys info)

- dependent on tchap-desktop https://github.com/tchapgouv/tchap-desktop/pull/142


src/vector/platform/tchap-desktop/TauriPlatform.ts
```diff
@@ -70,7 +70,7 @@ export default class TauriPlatform extends BasePlatform {
         this.protocol = "tchap";
 
         dis.register(onAction);
-        this.tauriSecureStorage = tauriSecureStorage;
+        this.tauriSecureStorage = new TauriSecureStorage("tchap.gouv.fr", this.ipc);
```
odelcroi: should this be dependent on the environment used ? I mean DEV is `tchap.incubateur.net`, is there a possible source of conflict if multiple tchap desktop are installed?

PR #1485, opened 2025-12-17
TCHAP(css): refactor tc css files into single folder

MarcWadai:
- regroup tchap css files into single tchap folder
- initialize pcss files into components.pcss instead of tsx component file


odelcroi:
looks good to me, nice cleansing

odelcroi (approved):

PR #1508, opened 2026-02-05
TCHAP(desktop): fix sometimes error on tauri ss initialization

MarcWadai:
It can happen : 
image

Looks like SDKConfig.get during tauriSecureStorage initialization is called but not yet available


odelcroi (approved):
solved my problem

